PR #90, opened 2022-11-04
fix: fix calculating fee percentiles

maksymar:
This PR fixes calculating fee percentiles by using 101 buckets (inclusive from 0 to 100).

Follow-up for [PR88](https://github.com/dfinity/bitcoin-canister/pull/88).
Current implementation is using incorrect number of percentile buckets (which is 100). 
But it should be 101, since we want to provide percentiles in the range [0, 100], inclusive.

### Test case n.1 (most common case): Number of values is higher than number of buckets.
If we have an input array of integer values [0, 1, 2, ... 10_000] we should expect the following percentiles:
- percentile[0] == 0
- percentile[50] == 500
- percentile[100] == 10_000

Note: all the results are integers as well.

### Test case n.2 (synthetic, easy to create): Number of values is lower than number of buckets.
If number of input values is lower than number of buckets, then calculated percentiles are fractional numbers, since many buckets are pointing to the space in between input values. It is easy to verify for example using `numpy.percentile(array, percentile)` in `Python`:

```py
assert np.percentile([1, 2], 50) == 1.5
```
 
Since we can not use fractional numbers we should stick to the lower bound element value, which is integer.

Below you can find some python verification code:

```py
import numpy as np

def expect(arr, q, expected):
    result = np.percentile(arr, q)
    if result == expected:
        print(f'[ OK ] {arr} {q} {expected}')
    else:
        print(f'[FAIL]: {arr} {q} {expected}!={result}')

# Input is larger than number of percentiles, integer results.
arr = range(0, 10_001)
expect(arr, 0, 0)
expect(arr, 9, 900)
expect(arr, 10, 1000)
expect(arr, 50, 5000)
expect(arr, 90, 9000)
expect(arr, 100, 10000)

# Input is smaller than number of percentiles, floating point results.
arr = [5, 4, 3, 2, 1]
expect(arr, 0, 1)
expect(arr, 25, 2.0)
expect(arr, 50, 3.0)
expect(arr, 75, 4.0)
expect(arr, 100, 5.0)

# Input is smaller than number of percentiles, floating point results.
arr = [33, 25]
expect(arr, 0, 25)
expect(arr, 20, 26.6)
expect(arr, 40, 28.2)
expect(arr, 60, 29.8)
expect(arr, 80, 31.4)
expect(arr, 100, 33)
```


maksymar:
> My suggestion is to use the standard **nearest-rank method**, **inclusive**, with the **extension of a 0th percentile**. Concretely, for any percentile `p` in `{0, 1, ..., 100}`, the 

@THLO PTAL at my implementation.

As a reference point I took the example from https://en.wikipedia.org/wiki/Percentile#The_nearest-rank_method
Turns out it reflects `inverted_cdf` method for `np.percentile`, so I used it as a verifier.

```py
import numpy as np

def custom_impl_percentile(arr, p):
    buckets = 101
    n = len(arr)
    if n < 100:
        x = n * (p-1) // (buckets-1)
    else:
        x = n * p // (buckets-1)
    i = max(0, min(x, n-1))
    return sorted(arr)[i]

def percentile(arr, p, expected):
    result = np.percentile(arr, p, method='inverted_cdf')
    if expected == result:
        print(f'[ OK ] percentile(arr, {p}) -> {expected}')
    else:
        print(f'[FAIL]: percentile(arr, {p}) -> {expected}!={result} (expected)')
    custom_impl_result = custom_impl_percentile(arr, p)
    if expected != custom_impl_result:
        print(f'[FAIL]: custom implementation failed for {p} {custom_impl_result}')

arr = [15, 20, 35, 40, 50]

percentile(arr, p=0, expected=15)
percentile(arr, p=20, expected=15)
percentile(arr, p=21, expected=20)
percentile(arr, p=40, expected=20)
percentile(arr, p=41, expected=35)
percentile(arr, p=60, expected=35)
percentile(arr, p=61, expected=40)
percentile(arr, p=80, expected=40)
percentile(arr, p=81, expected=50)
percentile(arr, p=100, expected=50)

arr = range(0, 101)
percentile(arr, p=0, expected=0)
percentile(arr, p=25, expected=25)
percentile(arr, p=50, expected=50)
percentile(arr, p=75, expected=75)
percentile(arr, p=100, expected=100)

arr = range(0, 1_001)
percentile(arr, p=0, expected=0)
percentile(arr, p=25, expected=250)
percentile(arr, p=50, expected=500)
percentile(arr, p=75, expected=750)
percentile(arr, p=100, expected=1_000)
```

Output:
```
percentile(arr, 0) -> 15
percentile(arr, 20) -> 15
percentile(arr, 21) -> 20
percentile(arr, 40) -> 20
percentile(arr, 41) -> 35
percentile(arr, 60) -> 35
percentile(arr, 61) -> 40
percentile(arr, 80) -> 40
percentile(arr, 81) -> 50
percentile(arr, 100) -> 50
```

THLO:
Thanks for the update! I think we're almost there.   :-)

There is still one discrepancy between the specification and the implementation:
The nearest-rank method returns the fee with **rank** p*(10,000/100) = p*100.
So, the median of the numbers [0, 1, 2, ..., 9999] should be the number with rank 5000, which is 4999.
In your implementation, the number with rank 5001, i.e., 5000, is returned instead.
At least that's what I got with my Python implementation of your code change. Let me know if I got it wrong!

maksymar:
@THLO thanks for the example! PTAL

Fixed the formula to match exactly [wiki](https://en.wikipedia.org/wiki/Percentile#The_nearest-rank_method):
- `ordinal_rank = ceil( (p/100) * n )`
- `index = limit(ordinal_rank-1, 0, n-1)`

ielashi:
@maksymar Would you also mind running `bash e2e-tests/profile.sh` and merge the changes to this PR so that we track the impact of the code changes on the performance?

canister/src/api/fee_percentiles.rs
```diff
@@ -125,23 +119,24 @@ fn get_tx_fee_per_byte(
     }
 }
 
-// Returns a requested number of percentile buckets from an initial vector of values.
-fn percentiles(mut values: Vec<u64>, buckets: u16) -> Vec<u64> {
+/// Compute percentiles of input values.
+///
+/// Returns 101 bucket to cover `[0, 100]` percentiles range.
```
THLO, line 124:
```suggestion
/// Returns 101 bucket to cover the percentiles range `[0, 100]`.
```

canister/src/api/fee_percentiles.rs
```diff
@@ -158,65 +153,69 @@ mod test {
...
+        assert_eq!(percentiles[0..21], [1; 21]);
+        assert_eq!(percentiles[21..41], [2; 20]);
+        assert_eq!(percentiles[41..61], [3; 20]);
+        assert_eq!(percentiles[61..81], [4; 20]);
+        assert_eq!(percentiles[81..101], [5; 20]);
     }
 
     #[test]
-    /// Given the input [1, 2, ..., 1000] and 100 buckets, the test ensures that the computed fees
+    /// Given the input [1, 2, ..., 1000] the test ensures that the computed fees
```
THLO, line 204:
```suggestion
    /// Given the input [1, 2, ..., 1000], the test ensures that the computed fees
```

THLO (approved):
Nice!

canister/src/api/fee_percentiles.rs
maksymar: Done.

canister/src/api/fee_percentiles.rs
maksymar: Done.

PR #95, opened 2022-11-17
feat: [EXC-1283] script to compute UTXOs.

ielashi:
Add a script for computing the UTXOs from a UTXO dump file.

### DIFF
--- a/bootstrap/state-builder/src/build_utxos.rs
+++ b/bootstrap/state-builder/src/build_utxos.rs
@@ -1,0 +1,149 @@
+//! A script for building the Bitcoin canister's UTXOs from a UTXO dump text file.
+//!
+//! Example run:
+//!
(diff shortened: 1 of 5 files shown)


bootstrap/state-builder/src/build_utxos.rs
```diff
@@ -0,0 +1,149 @@
...
+    Memory,
+};
+use std::{
+    fs::{create_dir_all, File},
+    io::{BufRead, BufReader, Write},
+    path::PathBuf,
+    str::FromStr,
+};
+
+const WASM_PAGE_SIZE: u64 = 65536;
```
dsarlis, line 26: I almost didn't recognise this without the 64 * 1024 form :P.

dsarlis (approved):

PR #98, opened 2022-11-21
chore: upgrade `stable-structures` to 0.3

ielashi:


canister/src/types.rs
```diff
@@ -445,7 +465,42 @@ impl Storable for (Height, OutPoint) {
...
+    }
+
+    fn from_bytes(bytes: Vec<u8>) -> Self {
+        Self::from(bytes)
+    }
+}
+
+impl BoundedStorable for BlockHeaderBlob {
+    fn max_size() -> u32 {
+        80
```
dsarlis, line 483: Might be worth explaining where we get this "magic" number from. Having a constant for it would be nice as well.

canister/src/types.rs
```diff
@@ -367,6 +373,14 @@ impl StableStructuresStorable for Address {
     }
 }
 
+impl BoundedStorable for Address {
+    fn max_size() -> u32 {
+        // The longest addresses are bech32 addresses, and a bech32 string can be at most 90 chars.
+        // See https://github.com/bitcoin/bips/blob/master/bip-0173.mediawiki
+        90
```
dsarlis, line 380: nit: Define a constant that represents this number.

dsarlis (approved):

PR #100, opened 2022-11-23
feat: [EXC-1286] Validate the bytes passed to the uploader canister.

ielashi:
To ensure the correctness of the bytes uploaded via the `uploader` canister, the `uploader` canister is now initialized with a list of the sha2 hashes of each chunk that'll be uploaded. The `uploader` then verifies the hash of each chunk prior to writing it into stable memory.

A script for compute the chunk hashes is also included.

### DIFF
--- a/bootstrap/uploader/src/compute_hashes.rs
+++ b/bootstrap/uploader/src/compute_hashes.rs
@@ -1,0 +1,35 @@
+//! A script for computing the hashes of the chunks of a file.
+//!
+//! The size of the chunk is hard-coded in `CHUNK_SIZE_IN_BYTES`.
+use clap::Parser;
+use std::{fs::File, io::Read, path::PathBuf};
+use uploader::*;
+
+#[derive(Parser, Debug)]
+struct Args {
+    /// The path of the file to compute its hashes.
+    #[clap(long, value_hint = clap::ValueHint::DirPath)]
+    file: PathBuf,
+}
+
+fn main() {
+    let args = Args::parse();
+
+    // Read file in chunks, and print the sha256 hash of each of these chunks.
+    let mut file = File::open(args.file).expect("opening file must succeed");
+    let mut chunk = vec![0; CHUNK_SIZE_IN_BYTES as usize];
+    loop {
+        let bytes_read = file.read(&mut chunk).unwrap();
+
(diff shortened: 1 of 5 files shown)


bootstrap/uploader/src/compute_hashes.rs
```diff
@@ -0,0 +1,35 @@
...
+        if bytes_read == 0 {
+            break;
+        }
+
+        let hash = sha256::digest(&chunk[0..bytes_read]);
+        println!("{}", hash);
+
+        if bytes_read < CHUNK_SIZE_IN_BYTES as usize {
+            break;
+        }
```
dsarlis, line 33: I suppose there's an edge case here if the file size is completely divisible by chunk size? Or in other words, we're sort of assuming that the last chunk would always be less than `CHUNK_SIZE_IN_BYTES` and the loop terminates. Perhaps, we need to calculate the number of chunks at the beginning and then ensure we read up to these many chunks?

dsarlis:
I took the liberty to fix some typos in the description, feel free to double check what I've changed.

bootstrap/uploader/src/compute_hashes.rs
ielashi: Ah, good catch. Will fix.

ielashi: @dsarlis  Actually, now that I am looking into this again, this is not the case. Even if the file size is divisible by the chunk size, the second termination condition won't succeed, but then we'll iterator through the loop again and the first termination condition will kick in.

dsarlis: Hmm, yeah indeed. I guess the structure of the code tripped me a little bit. But now that you point this out, do we even need the second condition then? It seems that only checking `bytes_read == 0` would be enough in any case (basically the second condition is sort of trying to optimize for not getting once more in the loop but I think it's actually adding confusion as evident by this convo).

dsarlis (approved):

bootstrap/uploader/src/compute_hashes.rs
ielashi: True, the second condition isn't useful. Removed.

PR #101, opened 2022-11-24
feat: implement `send_transaction` endpoint.

ielashi:
Adds a `send_transaction` endpoint to the bitcoin canister, which simply relays the request to the `bitcoin_send_transaction_internal` endpoint on the replica.


canister/src/api/send_transaction.rs
```diff
@@ -0,0 +1,52 @@
...
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    use crate::types::{Config, Fees, Network};
+    use ic_btc_types::NetworkInRequest;
+
+    #[async_std::test]
+    async fn charges_cycles() {
```
dsarlis, line 33:
```suggestion
    async fn send_transaction_charges_cycles() {
```

dsarlis (approved):

canister/src/api/send_transaction.rs
ielashi: The test names are already namespaced with the name of the module, so adding the module name is redundant:

```
test api::send_transaction::test::charges_cycles ... ok
```

PR #103, opened 2022-11-25
fix: load all supported addresses from the utxo dump file

ielashi:
When pre-computing the state we rely on a 3rd party tool that outputs a text dump of the UTXOs. Not all the addresses we support are emitted by this tool, so as a result we don't include all the addresses we support in our state.

This commit fixes this issue by adding a fallback mechanism. If parsing the address fails (e.g. it doesn't exist), we attempt to parse the script of the UTXO directly.

### DIFF
--- a/bootstrap/state-builder/src/build_balances.rs
+++ b/bootstrap/state-builder/src/build_balances.rs
@@ -60,13 +60,11 @@ fn main() {
         // we support, so if parsing the address itself fails, we try parsing the script directly.
         let address = if let Ok(address) = BitcoinAddress::from_str(address_str) {
             Some(address)
-        } else if let Some(address) = BitcoinAddress::from_script(
-            &Script::from(hex::decode(script).expect("script must be valid hex")),
-            args.network.into(),
-        ) {
-            Some(address)
         } else {
-            None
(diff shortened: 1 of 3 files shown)


bootstrap/state-builder/src/build_balances.rs
```diff
@@ -50,12 +50,26 @@ fn main() {
...
+        let address = if let Ok(address) = BitcoinAddress::from_str(address_str) {
+            Some(address)
+        } else if let Some(address) = BitcoinAddress::from_script(
+            &Script::from(hex::decode(script).expect("script must be valid hex")),
+            args.network.into(),
+        ) {
+            Some(address)
+        } else {
+            None
+        };
```
dsarlis, line 70: Should we maybe move this to a common location so that we reuse it in the two files?

dsarlis (approved):

bootstrap/state-builder/src/build_balances.rs
ielashi: That would be preferable, of course. In the interest of time, however, and given this is mostly one-off code, I'll keep it as-is and we can revisit it later if need be.

PR #106, opened 2022-11-25
Fix compute scripts

dfinity-berestovskyy:


bootstrap/step_4.sh
```diff
@@ -3,14 +3,16 @@
...
 # Generate the UTXO set.
 ~/go/bin/bitcoin-utxo-dump -db ./data/chainstate -o utxodump.csv -f "height,txid,vout,amount,type,address,script,coinbase,nsize"
 
 echo "Removing the headers from the file..."
 tail -n +2 utxodump.csv > utxodump.csv.tmp && mv utxodump.csv.tmp utxodump.csv
 
-echo "Sorting the file..."
-sort -n -o utxodump.csv utxodump.csv
+echo "Sorting the file, stopping at height ${HEIGHT}..."
+sort -n utxodump.csv | sed "/^${HEIGHT},/q" > utxodump.csv.tmp && mv utxodump.csv.tmp utxodump.csv
```
ielashi, line 15: What is the motivation behind this change? The UTXO file should contain UTXOs _exactly_ up to the height, and nothing beyond that. Is that not what you're observing?

ielashi (commented):
@dfinity-berestovskyy Thanks so much for going out of your way to write this up! I left one question.

bootstrap/step_4.sh
dfinity-berestovskyy: The `stopatheight` is not quite precise. Here is a bug report with some workaround: https://github.com/bitcoin/bitcoin/issues/13477

We seems doing this trick with `invalidateblock` at `step_2.sh`, but then as we use `bitcoin-utxo-dump`, it still dumps the full database it seems.

ielashi: When we invalidate, then the database becomes exactly at the height that we want, so `invalidateblock` is enough to handle the non-determinism of `stopatheight`.

ielashi: If you're seeing a different behavior, that means there's a bug somewhere else, but we definitely cannot filter heights here in this way, as that would result in an incorrect UTXO set.

PR #109, opened 2022-11-29
feat: enhancements to the `send_transaction` endpoint.

ielashi:
* Adds a metric to count number of requests.
* Does some basic parsing of the transaction as a litmus-test for its validity.
* Adds more coverage to `scenario-3` and adds it to CI

### DIFF
--- a/canister/src/api/metrics.rs
+++ b/canister/src/api/metrics.rs
@@ -92,7 +92,7 @@ fn encode_metrics(w: &mut MetricsEncoder<Vec<u8>>) -> std::io::Result<()> {
         w.encode_instruction_histogram(&state.metrics.get_current_fee_percentiles_total)?;
 
         w.encode_gauge(
-            "ins_send_transaction_count",
+            "send_transaction_count",
             state.metrics.send_transaction_count as f64,
             "The total number of (valid) requests to the send_transaction endpoint.",
         )?;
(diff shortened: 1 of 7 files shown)


canister/src/api/metrics.rs
```diff
@@ -91,6 +91,12 @@ fn encode_metrics(w: &mut MetricsEncoder<Vec<u8>>) -> std::io::Result<()> {
         w.encode_instruction_histogram(&state.metrics.get_balance_apply_unstable_blocks)?;
         w.encode_instruction_histogram(&state.metrics.get_current_fee_percentiles_total)?;
 
+        w.encode_gauge(
+            "ins_send_transaction_count",
```
dsarlis, line 95: What does "ins" stand for? Also, should we add a label to differentiate between valid and invalid requests? It might be useful to know both counters.

dsarlis (approved):
> Does same basic parsing of the transaction as a litmus-test for its validity.

I assume you meant "some basic parsing" otherwise the sentence doesn't parse in my head (pun intended).

canister/src/api/metrics.rs
ielashi: "ins" stands for instructions - which now that you mention it, shouldn't be a prefix for this endpoint since it isn't counting instructions. I'll remove that.

I wanted to originally have a counter for both valid and invalid requests, but we cannot actually count invalid requests since we panic in those cases.

dsarlis: Ah, that makes sense. It's fine then, the part of having invalid requests was a nit.

PR #115, opened 2022-12-01
feat: new fee structure

ielashi:
Updates the fee structure to make pricing more fair and more future-proof, per the recent benchmarking work.

### DIFF
--- a/canister/src/lib.rs
+++ b/canister/src/lib.rs
@@ -160,6 +160,18 @@ pub(crate) fn genesis_block(network: Network) -> Block {
 }
 
 pub(crate) fn charge_cycles(amount: u128) {
+    verify_has_enough_cycles(amount);
+
+    let amount: u64 = amount.try_into().expect("amount must be u64");
+    assert_eq!(
+        msg_cycles_accept(amount),
+        amount,
+        "Accepting cycles must succeed"
+    );
+}
+
+/// Panics if the request less than the amount of cycles given.
+pub(crate) fn verify_has_enough_cycles(amount: u128) {
     let amount: u64 = amount.try_into().expect("amount must be u64");
 
     if msg_cycles_available() < amount {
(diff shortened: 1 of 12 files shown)


canister/src/types.rs
```diff
@@ -54,10 +54,34 @@ impl Default for Config {
 
 #[derive(CandidType, Serialize, Deserialize, PartialEq, Eq, Debug, Clone, Default)]
 pub struct Fees {
-    pub get_utxos: u128,
+    /// The base fee to charge for all `get_utxos` requests.
+    pub get_utxos_base: u128,
```
ielashi, line 58: Note that these changes will require us to temporarily add some `serde[default]`s. I'll be adding those, along with hard-coding the actual fees we'll use for the launch, in a separate commit.

dsarlis: What about the `get_utxos` field you're removing? Isn't that a potential issue with candid?

canister/src/lib.rs
```diff
@@ -160,6 +160,18 @@ pub(crate) fn genesis_block(network: Network) -> Block {
...
+
+    let amount: u64 = amount.try_into().expect("amount must be u64");
+    assert_eq!(
+        msg_cycles_accept(amount),
+        amount,
+        "Accepting cycles must succeed"
+    );
+}
+
+/// Panics if the request less than the amount of cycles given.
```
dsarlis, line 173:
```suggestion
/// Panics if the request contains less than the amount of cycles given.
```

dsarlis (approved):

canister/src/types.rs
ielashi: It does break the `set_config` API, but that's a private endpoint so that isn't an issue.

dsarlis: I'm also not entirely sure that decoding an old state with the new struct would work. You tested specifically the upgrade of the bitcoin canister to this version on the testnet?

PR #116, opened 2022-12-01
chore: reduce max UTXOs per response to 1k.

ielashi:
This change is in line with the new pricing we're introducing.

### DIFF
--- a/canister/src/api/get_utxos.rs
+++ b/canister/src/api/get_utxos.rs
@@ -13,13 +13,13 @@ use std::str::FromStr;
 // `GetUtxosResponse`.
 //
 // Given the size of a `Utxo` is 48 bytes, this means that the size of a single
-// response can be ~500KiB (considering the size of remaining fields and
-// potential overhead for the candid serialization). This is still quite below
+// response can be ~50KiB (considering the size of remaining fields and potential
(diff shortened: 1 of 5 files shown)


canister/src/api/get_utxos.rs
```diff
@@ -13,13 +13,13 @@ use std::str::FromStr;
...
-// potential overhead for the candid serialization). This is still quite below
+// response can be ~50KiB (considering the size of remaining fields and potential
+// overhead for the candid serialization). This is still quite below
 // the max response payload size of 2MiB that the IC needs to respect.
-
+//
 // The value also conforms to the interface spec which requires that no more
-// than 100_000 `Utxo`s are returned in a single response.
-const MAX_UTXOS_PER_RESPONSE: usize = 10_000;
+// than 10_000 `Utxo`s are returned in a single response.
```
dsarlis, line 21: AFAICT the spec still mentions the 100_000 upper limit, so probably we shouldn't change this comment.

dsarlis (approved):

canister/src/api/get_utxos.rs
ielashi: There's a PR that changes this in review right now which will be merged soon.

PR #118, opened 2022-12-06
prettify cycles print

maksymar:
This PR prettifies cycles in error messages.

Before: `Received 1234567890 cycles. 999234567890 cycles are required.`
After: `Received 1_234_567_890 cycles. 999_234_567_890 cycles are required.`

### DIFF
--- a/canister/src/types.rs
+++ b/canister/src/types.rs
@@ -13,7 +13,14 @@ use ic_stable_structures::{BoundedStorable, Storable as StableStructuresStorable
 use serde::{Deserialize, Serialize};
 use serde_bytes::ByteBuf;
 use std::cell::RefCell;
-use std::{cmp::Ordering, convert::TryInto, str::FromStr};
+use std::{
+    cmp::Ordering,
+    convert::TryInto,
+    fmt,
+    ops::{Add, Mul},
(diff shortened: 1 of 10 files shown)


canister/src/types.rs
```diff
@@ -52,37 +59,82 @@ impl Default for Config {
...
+    fn mul(self, rhs: usize) -> Self {
+        Self(self.0.saturating_mul(rhs as u128))
+    }
+}
+
 #[derive(CandidType, Serialize, Deserialize, PartialEq, Eq, Debug, Clone, Default)]
 pub struct Fees {
     /// The base fee to charge for all `get_utxos` requests.
-    pub get_utxos_base: u128,
+    pub get_utxos_base: Cycles,
```
dsarlis, line 110: I'm not sure this change is fine from a backwards compatibility point of view. We'll need to double check whether we can successfully upgrade the canister to this version.

maksymar: That makes sense. I created another PR, simplified version with only adding underscores and no changing types.

https://github.com/dfinity/bitcoin-canister/pull/119

maksymar:
Closed in favour of #119

PR #119, opened 2022-12-06
feat: prettify cycles error message

maksymar:
This PR prettifies cycles in error messages.

Before: `Received 1234567890 cycles. 999234567890 cycles are required.`
After: `Received 1_234_567_890 cycles. 999_234_567_890 cycles are required.`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1222,6 +1222,7 @@ dependencies = [
  "serde",
  "serde_bytes",
  "tempfile",
+ "thousands",
 ]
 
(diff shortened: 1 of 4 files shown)


ielashi:
@maksymar @dsarlis This change is effectively a breaking change, as any canister that happens to be parsing this error message will now break. Are we sure we want to introduce breaking changes at this point?

maksymar:
I was not aware it's a breaking change. Created #123 to revert message change.

PR #120, opened 2022-12-07
feat: Add a flag to disable bitcoin apis

dsarlis:
This PR adds a flag that allows us to enable/disable access to the bitcoin apis provided by the canister. The flag is set to have a default value of `true` to not disrupt the current operations of the canister. It's also added in the canister's `Config` so that it can eventually be set through a proposal that triggers the `set_config` endpoint of the canister. For now, if we need to disable the apis, we can set the flag to `Disabled` in the canister's post upgrade hook.

### DIFF
--- a/canister/src/api/send_transaction.rs
+++ b/canister/src/api/send_transaction.rs
@@ -1,39 +1,45 @@
 use crate::{
-    charge_cycles, runtime, types::SendTransactionInternalRequest, verify_network, with_state,
-    with_state_mut,
+    charge_cycles, runtime,
+    types::{Flag, SendTransactionInternalRequest},
+    verify_network, with_state, with_state_mut,
 };
 use bitcoin::{consensus::Decodable, Transaction};
 use ic_btc_types::SendTransactionRequest;
 
 pub async fn send_transaction(request: SendTransactionRequest) {
-    verify_network(request.network.into());
+    match with_state(|s| s.api_access) {
+        Flag::Disabled => panic!("`bitcoin_send_transaction` is disabled."),
+        Flag::Enabled => {
+            verify_network(request.network.into());
 
-    charge_cycles(with_state(|s| {
-        s.fees.send_transaction_base
-            + s.fees.send_transaction_per_byte * request.transaction.len() as u128
-    }));
+            charge_cycles(with_state(|s| {
+                s.fees.send_transaction_base
+                    + s.fees.send_transaction_per_byte * request.transaction.len() as u128
+            }));
 
-    // Decode the transaction as a sanity check that it's valid.
-    let tx = Transaction::consensus_decode(request.transaction.as_slice())
-        .expect("Cannot decode transaction");
+            // Decode the transaction as a sanity check that it's valid.
+            let tx = Transaction::consensus_decode(request.transaction.as_slice())
+                .expect("Cannot decode transaction");
 
-    runtime::print(&format!("[send_transaction] Tx ID: {}", tx.txid()));
+            runtime::print(&format!("[send_transaction] Tx ID: {}", tx.txid()));
 
-    // Bump the counter for the number of (valid) requests received.
-    with_state_mut(|s| {
-        s.metrics.send_transaction_count += 1;
-    });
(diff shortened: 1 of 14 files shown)


canister/src/api/send_transaction.rs
```diff
@@ -1,39 +1,45 @@
...
-        SendTransactionInternalRequest {
-            network: request.network.into(),
-            transaction: request.transaction,
-        },
-    )
-    .await
-    .expect("Sending transaction bitcoin network must succeed");
+    match with_state(|s| s.api_access) {
+        Flag::Disabled => panic!("`bitcoin_send_transaction` is disabled."),
+        Flag::Enabled => {
```
ielashi, line 12: nit: instead of nesting each method in a match, how about adding a `verify_api_access()` method? Then the code would look like this:

```
verify_api_access();
verify_network(request.network.into());
...
```

canister/src/api/send_transaction.rs
```diff
@@ -1,39 +1,45 @@
...
-        with_state(|s| s.blocks_source),
-        SendTransactionInternalRequest {
-            network: request.network.into(),
-            transaction: request.transaction,
-        },
-    )
-    .await
-    .expect("Sending transaction bitcoin network must succeed");
+    match with_state(|s| s.api_access) {
+        Flag::Disabled => panic!("`bitcoin_send_transaction` is disabled."),
```
ielashi, line 11: nit: maybe returning the same error message from all endpoints (e.g. "API is disabled"), as opposed to a different message for every endpoint, would make it a bit easier for callers to parse this message and detect this state?

canister/src/api/set_config.rs
```diff
@@ -23,6 +23,10 @@ fn set_config_no_verification(request: SetConfigRequest) {
                     .expect("stability threshold too large"),
             );
         }
+
+        if let Some(api_access) = request.api_access {
```
ielashi, line 27: Can we add a test case for this? See below for examples on how we test the other configurations.

ielashi (approved):
@dsarlis Thanks! Can we also run `bash e2e-tests/profile.sh` to include the performance changes? I know it's negligible, but it's good to have it documented in the PR as well.

canister/src/api/send_transaction.rs
dsarlis: I think if I implement your other suggestion above, it'll basically have to be a generic error message (or at least it's a lot easier that way).

dsarlis:
@ielashi I believe I took care of your comments. Feel free to double check. I was also thinking it would be a good idea to confirm we can upgrade the canister to this version (I think we should be able to) on the testnet we have it running. I couldn't do it myself because I'm not the controller of that canister. If you can do it as a quick check, I'd appreciate it.

ielashi:
> @ielashi I believe I took care of your comments. Feel free to double check. I was also thinking it would be a good idea to confirm we can upgrade the canister to this version (I think we should be able to) on the testnet we have it running. I couldn't do it myself because I'm not the controller of that canister. If you can do it as a quick check, I'd appreciate it.

Done. It's actually quite easy to test locally. I checked out the `release/2022-12-02` branch, deployed the canister, then switched to your branch, then upgraded the canister. I think it'd be great to have that check as part of CI. I created https://dfinity.atlassian.net/browse/EXC-1305 to track this.

dsarlis:
> > @ielashi I believe I took care of your comments. Feel free to double check. I was also thinking it would be a good idea to confirm we can upgrade the canister to this version (I think we should be able to) on the testnet we have it running. I couldn't do it myself because I'm not the controller of that canister. If you can do it as a quick check, I'd appreciate it.
> 
> Done. It's actually quite easy to test locally. I checked out the `release/2022-12-02` branch, deployed the canister, then switched to your branch, then upgraded the canister. I think it'd be great to have that check as part of CI. I created https://dfinity.atlassian.net/browse/EXC-1305 to track this.

Ah right. I guess I wrongly assumed that I need to have the full state on the bitcoin canister but yeah it doesn't need to have the full state for this type of checks. Thanks for checking!

PR #124, opened 2022-12-13
chore: unify canister build scripts.

ielashi:
We had three different scripts for building canisters. This commit deletes two of them and only keeps one that's used for all the builds.


dsarlis (approved):
I think we probably need to update `Dockerfile` too to use the new name for the bitcoin canister. Which probably also shows that we should add it in the CI pipelines too to ensure it doesn't get broken.

ielashi:
> I think we probably need to update `Dockerfile` too to use the new name for the bitcoin canister. Which probably also shows that we should add it in the CI pipelines too to ensure it doesn't get broken.

Done, good catch. I'm a bit hesitant to add the docker build to CI as it would slow down CI considerably, and these codepaths would rarely get updated moving forward. I'd suggest we wait and see. If docker breakages keep coming up, then we add it to CI.

dsarlis:
> > I think we probably need to update `Dockerfile` too to use the new name for the bitcoin canister. Which probably also shows that we should add it in the CI pipelines too to ensure it doesn't get broken.
> 
> Done, good catch. I'm a bit hesitant to add the docker build to CI as it would slow down CI considerably, and these codepaths would rarely get updated moving forward. I'd suggest we wait and see. If docker breakages keep coming up, then we add it to CI.

I believe this is one of the things that's just meant to be broken if it's not on CI. It would add some slow down indeed, but it's something like 15-20 minutes for a clean build (at least based on my experience) which I think is still very much acceptable as far as CI times go (and the review speed which is usually not faster than a couple hours anyway).

PR #126, opened 2022-12-14
feat: use difficulty-based stability threshold for fork resolution

dragoljub-duric:


canister/src/unstable_blocks.rs
```diff
@@ -170,31 +178,32 @@ pub fn get_chain_with_tip<'a, 'b>(
...
-    let mut depths: Vec<_> = blocks
+    // Compute the normalized weight of all the children.
+    let network = blocks.network;
+    let mut weights: Vec<_> = blocks
         .tree
         .children
         .iter()
         .enumerate()
-        .map(|(idx, child)| (blocktree::depth(child), idx))
+        .map(|(idx, child)| (blocktree::get_normalized_weight(child, network), idx))
```
ielashi, line 188: This is incorrect. What this code is doing is summing the difficulties of each child, then normalizing those difficulties with the root of each child. What we want is to normalize with the `blocks.tree.root`. The fix here would be to call `get_max_weight` here, then in a later step in this method divide the difficulty of each one by `blocks.tree.root`.

I'd also add a test case to ensure that we are normalizing with `blocks.tree.root`.

canister/src/types.rs
```diff
@@ -119,11 +123,31 @@ impl Block {
...
+
+    #[cfg(test)]
+    pub fn get_difficulty(&self, network: BitcoinNetwork) -> u64 {
+        if let Some(difficulty) = self.mock_dificulty {
+            difficulty
+        } else {
+            self.header().difficulty(network)
+        }
+    }
+
```
ielashi, line 139: A few nits:

1. I'd rename this method to `difficulty()` to make the naming consistent with other methods on this struct (e.g. `header()`).
2. Please use the `crate::types::Network` as the parameter. We only use the `BitcoinNetwork` internally within methods, but we don't want it to be part of method signatures when possible (there are easy conversion methods available between the two types).
3. This code can be written more succinctly like this (untested):

```
    pub fn difficulty(&self, network: BitcoinNetwork) -> u64 {
        #[cfg(test)]
        if let Some(difficulty) = self.mock_dificulty {
            difficulty
        }
         
         self.header().difficulty(network)
    }
```

canister/src/types.rs
```diff
@@ -93,6 +93,8 @@ pub struct Fees {
 pub struct Block {
     block: BitcoinBlock,
     transactions: Vec<Transaction>,
+    #[cfg(test)]
+    mock_dificulty: Option<u64>,
```
ielashi, line 97:
```suggestion
    mock_difficulty: Option<u64>,
```

canister/src/blocktree.rs
```diff
@@ -185,7 +187,23 @@ fn get_chain_with_tip_reverse<'a, 'b>(
     None
 }
 
+fn get_max_weight(tree: &BlockTree, network: BitcoinNetwork) -> u128 {
```
ielashi, line 190:
```suggestion
fn difficulty_based_depth(tree: &BlockTree, network: BitcoinNetwork) -> u128 {
```
The terminology in the [wiki](https://wiki.internetcomputer.org/wiki/Bitcoin_integration) has been updated to not use the term "weight" anymore, as that conflicts with another concept in Bitcoin.

canister/src/unstable_blocks.rs
dragoljub-duric: Maybe in that case it is better to avoid division and just multiply stability treshlod with `blocks.tree.root.difficulty()`, because it will give the same result.

THLO: Yes, multiplying the threshold with the root difficulty would work as well (without the use of division).

canister/src/blocktree.rs
```diff
@@ -185,7 +185,18 @@ fn get_chain_with_tip_reverse<'a, 'b>(
...
+    let mut res: u128 = 0;
+    for child in tree.children.iter() {
+        res = std::cmp::max(res, difficulty_based_depth(child));
+    }
+    res += tree.root.difficulty() as u128;
+    res
+}
+
 /// Returns the depth of the tree.
+#[cfg(test)]
```
ielashi, line 199: Do we still need this method? If not, consider deleting it.

canister/src/unstable_blocks.rs
```diff
@@ -170,32 +170,36 @@ pub fn get_chain_with_tip<'a, 'b>(
...
         .enumerate()
-        .map(|(idx, child)| (blocktree::depth(child), idx))
+        .map(|(idx, child)| (blocktree::difficulty_based_depth(child), idx))
         .collect();
 
     // Sort by depth.
     depths.sort_by_key(|(depth, _child_idx)| *depth);
 
+    let root_difficulty = std::cmp::max(1, blocks.tree.root.difficulty()) as u128;
+
```
ielashi, line 186: Given that we are no longer dividing, do we need the `max(1, ..)` logic at this point?

canister/src/unstable_blocks.rs
```diff
@@ -244,43 +252,160 @@ mod test {
...
+        push(&mut forest, &utxos, block_1.clone()).unwrap();
+        push(&mut forest, &utxos, block_2).unwrap();
+        assert_eq!(peek(&forest), None);
+        assert_eq!(pop(&mut forest), None);
+
+        push(&mut forest, &utxos, block_3).unwrap();
+        // Block 0 (the anchor) now has one stable child (Block 1).
+        // Block 0 should be returned when calling `pop`.
+        assert_eq!(peek(&forest), Some(&block_0));
```
ielashi, line 285: nit: I think it would be useful here and elsewhere in tests to explain why block 1 has now become stable. It's as obvious anymore with the additional complexity we're introducing with `difficulty`.

canister/src/validation.rs
```diff
@@ -76,12 +76,21 @@ mod test {
...
+
+            let stability_threshold = stability_threshold_start + 1;
+
+            let blocks_start = build_chain(network, num_blocks, num_transactions_in_block);
+
+            let mut blocks = vec![];
+
+            for block in blocks_start.into_iter(){
+                blocks.push(block.with_mock_dificulty(1));
+            }
```
ielashi, line 93: I think you can do this (untested), which is a bit more "Rustic":

```suggestion
            let blocks: Vec<_> = build_chain(network, num_blocks, num_transactions_in_block)
              .into_iter().map(|b| b.with_mock_difficulty(1)).collect();
```

canister/src/validation.rs
```diff
@@ -76,12 +76,21 @@ mod test {
...
         fn validation_context(
-            stability_threshold in 1..150u32,
+            stability_threshold_start in 1..150u32,
             num_blocks in 1..250u32,
         ) {
             let num_transactions_in_block = 1;
             let network = Network::Regtest;
-            let blocks = build_chain(network, num_blocks, num_transactions_in_block);
+
+            let stability_threshold = stability_threshold_start + 1;
```
ielashi, line 85: Why do we need to increment the stability threshold by one?

ielashi (commented):
Awesome work @dragoljub-duric ! I hope you don't mind me nitpicking a bit :)

canister/src/unstable_blocks.rs
dragoljub-duric: I mean we can remove max(), but in that case, if blocks.tree.root.difficulty() is zero, any block will be stable.

canister/src/validation.rs
dragoljub-duric: If we have the tree with only the root with difficulty 1, `fn depth()` will return 0, while `fn difficulty_based_depth()` will return 1. The same follows in all other cases. Hence to make the same behavior we need to increment the stability threshold by one.

canister/src/unstable_blocks.rs
ielashi: I think that's fine, and expected, but maybe @THLO can confirm as well.

THLO: As discussed with @ielashi offline, the [minimum difficulty](https://en.bitcoin.it/wiki/Difficulty) should be 1, so we shouldn't run into the situation where we need to deal with a difficulty of 0.

canister/src/validation.rs
THLO: The specification says that depth is a function of the successors, so `difficulty_based_depth()` should also return 0 because there are no successors, right?

dragoljub-duric: Both `depth()` and `difficulty_based_depth()` are called on the children of the root. If the root does not have children then they will never be called. But in case the child of the root has difficulty 1, and does not have any successors `depth()` will return 0, while `difficulty_based_depth()` will return 1.

THLO: I don't understand why it returns 1. If this child has no successors, the sum over the difficulties must be 0. If we normalize, we get 0/1 = 0.

dragoljub-duric: But `child`'s difficulty is 1, hence the path with the max difficulty in the subtree that has `child` as the root has difficulty 1.

canister/src/blocktree.rs
```diff
@@ -185,19 +185,14 @@ fn get_chain_with_tip_reverse<'a, 'b>(
...
-pub fn depth(block_tree: &BlockTree) -> u32 {
-    if block_tree.children.is_empty() {
-        return 0;
-    }
-
-    let mut max_child_depth = 0;
-
-    for child in block_tree.children.iter() {
-        max_child_depth = std::cmp::max(1 + depth(child), max_child_depth);
+// Returns depth based on the difficulty.
```
ielashi, line 188:
```suggestion
// Returns the maximum sum of block difficulties from the root to a leaf inclusive.
```

canister/src/blocktree.rs
```diff
@@ -185,19 +185,14 @@ fn get_chain_with_tip_reverse<'a, 'b>(
...
+pub fn difficulty_based_depth(tree: &BlockTree, network: Network) -> u128 {
```
ielashi, line 189: nit: can we add a unit test for this method?

canister/src/lib.rs
```diff
@@ -144,7 +144,14 @@ pub fn post_upgrade() {
...
 
     // Deserialize and set the state.
-    let state = ciborium::de::from_reader(&*state_bytes).expect("failed to decode state");
+    let mut state: State =
+        ciborium::de::from_reader(&*state_bytes).expect("failed to decode state");
+
+    // to be removed after upgrade
+    let network = state.network();
+
+    state.unstable_blocks = state.unstable_blocks.with_network(network); //EXC-1310
```
ielashi, line 153:
```suggestion
    // TODO(EXC-1310): to be removed after upgrade
    let network = state.network();
    state.unstable_blocks = state.unstable_blocks.with_network(network);
```

canister/src/types.rs
```diff
@@ -119,6 +123,14 @@ impl Block {
...
     }
 
+    pub fn difficulty(&self, network: Network) -> u64 {
+        #[cfg(test)]
+        if let Some(difficulty) = self.mock_difficulty {
+            return difficulty;
+        }
+
+        Self::target_difficulty(network, self.header().target())
+    }
```
ielashi, line 133: nit: a new line between the methods.

canister/src/unstable_blocks.rs
```diff
@@ -18,10 +18,16 @@ pub struct UnstableBlocks {
...
+    network: Option<Network>, // EXC-1310
 }
 
 impl UnstableBlocks {
-    pub fn new(utxos: &UtxoSet, stability_threshold: u32, anchor: Block) -> Self {
+    pub fn new(
+        utxos: &UtxoSet,
+        stability_threshold: u32,
+        anchor: Block,
+        network: Option<Network>, // Optional just for the upgrade, will be refactored after. EXC-1310
```
ielashi, line 29:
```suggestion
        network: Option<Network>, // TODO(EXC-1310): Optional just for the upgrade, will be refactored after.
```

canister/src/unstable_blocks.rs
```diff
@@ -59,6 +66,16 @@ impl UnstableBlocks {
     pub fn set_stability_threshold(&mut self, stability_threshold: u32) {
         self.stability_threshold = stability_threshold;
     }
+
+    fn get_network(&self) -> Option<Network> {
+        self.network
+    }
+
+    // temporary method will be removed after an upgrade. EXC-1310
```
ielashi, line 74:
```suggestion
    // TODO(EXC-1310): temporary method will be removed after an upgrade.
```

canister/src/unstable_blocks.rs
```diff
@@ -170,32 +187,42 @@ pub fn get_chain_with_tip<'a, 'b>(
 
 // Returns the index of the `anchor`'s stable child if it exists.
 fn get_stable_child(blocks: &UnstableBlocks) -> Option<usize> {
-    // Compute the depth of all the children.
+    // Compute the difficulty based depth of all the children.
+    let network = if let Some(network) = blocks.get_network() {
+        network
+    } else {
+        panic!("Network should be defined.") // EXC-1310
+    };
```
ielashi, line 195:
```suggestion
    let network = self.get_network().expect("Network should be defined.");
```

canister/src/unstable_blocks.rs
```diff
@@ -218,20 +245,25 @@ mod test {
...
+        let block_0 = BlockBuilder::genesis().build().with_mock_difficulty(1);
+        let block_1 = BlockBuilder::with_prev_header(block_0.header())
+            .build()
+            .with_mock_difficulty(1);
+        let block_2 = BlockBuilder::with_prev_header(block_1.header())
+            .build()
+            .with_mock_difficulty(1);
+        let network = Network::Mainnet;
+        let utxos = UtxoSet::new(network);
+        let mut forest = UnstableBlocks::new(&utxos, 2, block_0.clone(), Some(network));
```
ielashi, line 266: I don't think you need to mock the difficulty here. Instead, you can set the network to be `Network::Regtest`, and the difficulty of all these blocks will be 1.

canister/src/unstable_blocks.rs
```diff
@@ -244,43 +276,177 @@ mod test {
...
+        push(&mut forest, &utxos, block_1.clone()).unwrap();
+        push(&mut forest, &utxos, block_2).unwrap();
+        assert_eq!(peek(&forest), None);
+        assert_eq!(pop(&mut forest), None);
+
+        push(&mut forest, &utxos, block_3).unwrap();
+        // block_0 (the anchor) now has stable child block_1. Because block_1's
+        // difficulty_based_depth is 15 + 20 + 110 = 145 is greater than
+        // normalized_stability_threshold is 20 * 7 = 140 and it does not have
+        // any siblings. Hence, block_0 should be returned when calling `pop`.
```
ielashi, line 311: I suggest also documenting these statements with code when possible. For example, the following can help document the fact that block1's d.b. stability threshold is 145:

```
assert_eq!(crate::blocktree::difficulty_based_depth(&forest.tree.children[0], network), 145);
```

(ditto elsewhere)

canister/src/types.rs
```diff
@@ -128,12 +140,15 @@ impl Block {
...
-    #[allow(dead_code)]
     fn target_difficulty(network: Network, target: Uint256) -> u64 {
         (ic_btc_validation::max_target(&network.into()) / target).low_u64()
     }
+
+    #[cfg(test)]
+    pub fn with_mock_difficulty(mut self, mock_difficulty: u64) -> Self {
+        self.mock_difficulty = Some(mock_difficulty);
+        self
+    }
```
ielashi, line 151: nit: I think this method belongs to the `BlockBuilder` in `test_utils.rs`. I'd make the `mock_difficulty` field public, and move this method into the `BlockBuilder`.

canister/src/validation.rs
ielashi: (unresolving) I'm still not clear on why we're touching this test. I don't see what test case fails by not incrementing the stability threshold by one, nor is it necessary anymore to mock the difficulty of the blocks below.

dragoljub-duric: You are right, we do not need it now, it is changed.

canister/src/types.rs
dragoljub-duric: Rust fmt will remove the new lines.

canister/src/unstable_blocks.rs
dragoljub-duric: Good point, it is updated.

canister/src/types.rs
dragoljub-duric: I guess it is unnecessary, and we will need to change a lot of code since we are not always using this method after BlockBuilder.

canister/src/blocktree.rs
dragoljub-duric: It is done.

canister/src/unstable_blocks.rs
dragoljub-duric: It is done, please give it a look.

canister/src/api/fee_percentiles.rs
```diff
@@ -419,8 +419,11 @@ mod test {
     fn get_current_fee_percentiles_from_utxos() {
         let number_of_blocks = 5;
         let number_of_transactions = 10_000;
-        let blocks = generate_blocks(10_000, number_of_blocks);
-        let stability_threshold = 1;
+        let blocks = generate_blocks(10_000, number_of_blocks)
```
ielashi, line 422: Do we need this change (and similarly below)?

canister/src/types.rs
ielashi: I don't have a strong opinion, but does it really change a lot of code? This method is called in ~10 places, and if we move it to the BlockBuilder, the updates we need to make in these places are trivial and would only take a few minutes right? If yes, then I think it's worth it to keep our block builder methods all in one place.

ielashi (approved):
LGTM :)

canister/src/api/fee_percentiles.rs
dragoljub-duric: I think so, I remember those tests were failing without it.

PR #128, opened 2022-12-16
chore: optimize computing next difficulty

ielashi:
In the block validation code, computing the next difficulty traverses back 2,016 blocks linearly. Given that these code paths will be using stable structures, this is quite wasteful. The interface of `HeaderStore` has been modified to allow fetching a block using its height, so that traversing 2,016 blocks is now done in one single call.

### DIFF
--- a/validation/src/header.rs
+++ b/validation/src/header.rs
@@ -38,7 +38,6 @@ pub trait HeaderStore {
     /// Returns the header at the given height.
     fn get_with_height(&self, height: u32) -> Option<BlockHeader>;
 
-
     /// Returns the height of the tip that the new header will extend.
     fn height(&self) -> u32;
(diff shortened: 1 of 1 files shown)


dsarlis (approved):
Can you add a short description explaining what exactly we enhance? I believe there are no functional changes but it's also not very clear to me what has been enhanced. I'm assuming it might also be setting the stage for something else coming up in a follow up?

ielashi:
> Can you add a short description explaining what exactly we enhance? I believe there are no functional changes but it's also not very clear to me what has been enhanced. I'm assuming it might also be setting the stage for something else coming up in a follow up?

I agree, it wasn't not clear at all. Done. Updated the PR title and description.

PR #129, opened 2022-12-19
feat: block header validation

ielashi:
Adds validation to the block headers. This code has been running on a testnet with the full mainnet state
to ensure that we don't have false negatives.

### DIFF
--- a/canister/src/api/fee_percentiles.rs
+++ b/canister/src/api/fee_percentiles.rs
@@ -392,13 +392,13 @@ mod test {
         });
     }
 
-    #[test]
+    /*#[test]
     fn get_current_fee_percentiles_big_input() {
         let number_of_blocks = 1_000;
         let initial_balance = 500_500; // number_of_blocks * (number_of_blocks + 1) / 2
         let blocks = generate_blocks(initial_balance, number_of_blocks);
         let stability_threshold = blocks.len() as u128;
-        init_state(blocks, stability_threshold);
+        init_state(blocks, 20);
 
         with_state_mut(|state| {
             let main_chain = unstable_blocks::get_main_chain(&state.unstable_blocks).into_chain();
@@ -426,7 +426,7 @@ mod test {
             assert_eq!(percentiles[61..81], [8386; 20]);
             assert_eq!(percentiles[81..101], [8394; 20]);
         });
-    }
+    }*/
 
     #[test]
     fn get_current_fee_percentiles_no_transactions() {
--- a/canister/src/api/fee_percentiles.rs
+++ b/canister/src/api/fee_percentiles.rs
@@ -392,42 +392,6 @@ mod test {
         });
     }
 
-    /*#[test]
-    fn get_current_fee_percentiles_big_input() {
-        let number_of_blocks = 1_000;
-        let initial_balance = 500_500; // number_of_blocks * (number_of_blocks + 1) / 2
-        let blocks = generate_blocks(initial_balance, number_of_blocks);
-        let stability_threshold = blocks.len() as u128;
-        init_state(blocks, 20);
-
-        with_state_mut(|state| {
-            let main_chain = unstable_blocks::get_main_chain(&state.unstable_blocks).into_chain();
-
-            let number_of_transactions = 5;
-            let fees = get_fees_per_byte(
-                main_chain.clone(),
-                &state.unstable_blocks,
-                number_of_transactions,
-            );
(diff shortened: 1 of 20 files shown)


canister/src/api/fee_percentiles.rs
```diff
@@ -392,42 +392,6 @@ mod test {
         });
     }
 
-    #[test]
-    fn get_current_fee_percentiles_big_input() {
-        let number_of_blocks = 1_000;
-        let initial_balance = 500_500; // number_of_blocks * (number_of_blocks + 1) / 2
-        let blocks = generate_blocks(initial_balance, number_of_blocks);
-        let stability_threshold = blocks.len() as u128;
-        init_state(blocks, stability_threshold);
```
ielashi, line 401: I deleted this test because with block header validation its runtime jumped from ~2 seconds to about 3 minutes. The reason it's slow is because it's using a stability threshold of > 500k, whereas we don't intend to have a stability threshold that's higher than 144 at this time. The code path can be optimized, but given that it's not something we intend to support, and given that this test (AFAICT) doesn't add extra coverage, I think deleting it would be the simplest. @maksymar WDYT?

canister/src/api/get_utxos.rs
```diff
@@ -913,50 +913,49 @@ mod test {
 
     #[test]
     fn get_utxos_does_not_include_other_addresses() {
-        for network in [Network::Mainnet, Network::Testnet, Network::Regtest].iter() {
```
ielashi, line 916: We can no longer create fake testnet and mainnet blocks as these fail the validation, so this test (and others) are now only generating regtest blocks.

canister/src/validation.rs
```diff
@@ -0,0 +1,57 @@
...
+use bitcoin::BlockHeader;
+use ic_btc_validation::HeaderStore;
+
+/// A structure passed to the validation crate to validate a specific block header.
+pub struct ValidationContext<'a> {
+    state: &'a State,
+    chain: Vec<&'a Block>,
+}
+
+impl<'a> ValidationContext<'a> {
```
ielashi, line 11: Note that there are no unit tests in this file, and that's because the current test suite already covers this code very well, so I didn't see a need to add additional tests, but I can add them if requested.

validation/src/header.rs
```diff
@@ -208,19 +215,21 @@ fn compute_next_difficulty(
...
-    }
-    // last_adjustment_header is the last header with height multiple of 2016
-    let last_adjustment_header = current_header;
+    // Computing the `last_adjustment_header`.
+    // `last_adjustment_header` is the last header with height multiple of 2016
+    let last_adjustment_height = if height < DIFFICULTY_ADJUSTMENT_INTERVAL {
+        0
+    } else {
+        height - DIFFICULTY_ADJUSTMENT_INTERVAL
+    };
```
ielashi, line 229: This is an optimization from the previous code. Rather than backtracking 2016 blocks, we simply fetch that block directly.

canister/src/validation.rs
dsarlis: I think ideally we'd have some tests that cover this module in isolation as it's used as a building block for the validation but I don't feel so strongly to block the PR on it.

canister/src/validation.rs
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    fn height(&self) -> u32 {
+        // The `next_height` method returns the height of the UTXOs + 1, so we
+        // subtract 1 to account for that.
+        self.state.utxos.next_height() + self.chain.len() as u32 - 1
+    }
+
+    fn get_with_height(&self, height: u32) -> Option<BlockHeader> {
+        if height < self.state.utxos.next_height() {
```
dsarlis, line 49: Maybe a short comment would help prepare the reader for the conditions that follow? Especially because we check against the `next_height` from stable utxos first and then against `self.height` and it took me a moment to confirm the correlation.

dsarlis (approved):
Did you intend to have this be based on https://github.com/dfinity/bitcoin-canister/pull/128? Some changes seem to be coming from that one actually. Or is this one supposed to supersede the previous one?

canister/src/validation.rs
ielashi: Done.

ielashi:
> Did you intend to have this be based on #128? Some changes seem to be coming from that one actually. Or is this one supposed to supersede the previous one?

Yes, correct. I merged #128 so this should now be resolved.

PR #131, opened 2022-12-21
feat: compute block difficulty

ielashi:
Computing the block difficulty is needed for computing the difficulty-based stability threshold that will be introduced in an upcoming commit.

We are not relying on the [difficulty](https://docs.rs/bitcoin/latest/bitcoin/blockdata/block/struct.BlockHeader.html#method.difficulty) method in the bitcoin crate, as it produces incorrect results for testnet and regtest.

### DIFF
--- a/canister/src/types.rs
+++ b/canister/src/types.rs
@@ -131,8 +131,8 @@ impl Block {
     //
     // NOTE: This is dead code temporarily and will be used in an upcoming PR.
     #[allow(dead_code)]
-    fn compute_difficulty(network: Network, block_target: Uint256) -> u64 {
-        (ic_btc_validation::max_target(&network.into()) / block_target).low_u64()
+    fn target_difficulty(network: Network, target: Uint256) -> u64 {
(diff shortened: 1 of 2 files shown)


canister/src/types.rs
```diff
@@ -977,3 +987,67 @@ fn address_handles_script_edge_case() {
...
+    );
+
+    // Mainnet block 768362.
+    // Data pulled from https://www.blockchain.com/explorer/blocks/btc/768362
+    assert_eq!(
+        Block::target_difficulty(
+            Network::Mainnet,
+            bitcoin::BlockHeader::u256_from_compact_target(386397584)
+        ),
+        35364065900457
```
dsarlis, line 1009: nit: Can we use underscores in the numbers for a bit better readability?

dsarlis (approved):

canister/src/types.rs
ielashi: Done.

THLO (approved):

dragoljub-duric (approved):

PR #137, opened 2023-01-11
Verify upgradability in CI

dragoljub-duric:
Closes [EXC-1305](https://dfinity.atlassian.net/browse/EXC-1305).

### DIFF
--- a/e2e-tests/verify-upgradability.sh
+++ b/e2e-tests/verify-upgradability.sh
@@ -39,7 +39,7 @@ dfx deploy --no-wallet bitcoin --argument "(record {
  api_access = variant { enabled }
 })"
 
-dfx canister stop --all
+dfx canister stop bitcoin
 
 # Move to the current branch
 git checkout "${CURRENT_BRANCH}"
--- a/e2e-tests/verify-upgradability.sh
+++ b/e2e-tests/verify-upgradability.sh
@@ -58,5 +58,26 @@ dfx deploy --no-wallet bitcoin --argument "(record {
 })"
 
 dfx canister start bitcoin
+dfx canister stop bitcoin
+
+dfx deploy --upgrade-unchanged bitcoin --argument "(record { 
+ stability_threshold = 2;
+ network = variant { regtest };
+ blocks_source = principal \"$(dfx canister id "${MANAGEMENT_CANISTER}")\";
+ fees = record { 
+    get_utxos_base = 0; 
+    get_utxos_cycles_per_ten_instructions = 0; 
+    get_utxos_maximum = 0; get_balance = 0; 
+    get_balance_maximum = 0; 
+    get_current_fee_percentiles = 0; 
+    get_current_fee_percentiles_maximum = 0;  
+    send_transaction_base =0; 
+    send_transaction_per_byte = 0; 
+ }; 
+ syncing = variant { enabled }; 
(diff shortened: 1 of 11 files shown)


e2e-tests/verify-upgradability.sh
```diff
@@ -0,0 +1,65 @@
...
+
+# Get newest release download url
+NEWEST_RELEASE="$(curl -s https://api.github.com/repos/dfinity/bitcoin-canister/releases/latest | grep "browser_download_url" | awk '{ print $2 }' | sed 's/,$//' | sed 's/"//g')"
+MANAGEMENT_CANISTER="aaaaa-aa"
+
+
+# Get newest release
+wget -O newest_release.wasm.gz "${NEWEST_RELEASE}"
+
+gunzip newest_release.wasm.gz
```
ielashi, line 15: I don't think you need to unzip, as the replica does this for you automatically (i.e. you can install the wasm.gz file directly).

e2e-tests/verify-upgradability.sh
```diff
@@ -0,0 +1,65 @@
...
+    send_transaction_per_byte = 0; 
+ }; 
+ syncing = variant { enabled }; 
+ api_access = variant { enabled }
+})"
+
+dfx canister stop bitcoin-release
+
+# replace from bitcoin-release to bitcoin in .dfx/local/canister_ids.json
+# hence, the upgraded canister has the same CanisteID
```
ielashi, line 41:
```suggestion
# hence, the upgraded canister has the same canister ID 
```

e2e-tests/verify-upgradability.sh
```diff
@@ -0,0 +1,65 @@
...
+    get_current_fee_percentiles_maximum = 0;  
+    send_transaction_base =0; 
+    send_transaction_per_byte = 0; 
+ }; 
+ syncing = variant { enabled }; 
+ api_access = variant { enabled }
+})"
+
+dfx canister start bitcoin
+
```
ielashi, line 64: After upgrading to the current commit, I would upgrade the canister again to the same commit so that we can also test the current commit's `pre_upgrade` hook. You can do something like this:

```
dfx canister stop bitcoin
dfx deploy --upgrade-unchanged bitcoin
dfx canister start bitcoin
```

dragoljub-duric: Done.

ielashi:
A couple of things:

1. I don't see the "verify-upgradability" test running the CI checks of this PR. Are you sure it was added correctly?
2. When I run `bash ./e2e-tests/verify-upgradability.sh` I get an error:

```
+ sed -i s/bitcoin-release/bitcoin/ ./.dfx/local/canister_ids.json
sed: 1: "./.dfx/local/canister_i ...": invalid command code .
```

ielashi:
Also, the error above aside, I _think_ this script will break if I am not in the root directory of the repo. We can easily make the script resilient to the path where it's called from by doing something like [this](https://github.com/dfinity/bitcoin-canister/blob/master/scripts/build-canister.sh#L9), where the script temporarily moves to its own directory at the beginning.

dragoljub-duric:
> I run it with the same command (on a mac), and it worked, I do not know how why that happens.

ielashi (approved):

PR #139, opened 2023-01-18
feat: Use stability count for counting confirmations.

dragoljub-duric:
Closes [EXC-1319](https://dfinity.atlassian.net/browse/EXC-1319).

### DIFF
--- a/canister/src/api/get_utxos.rs
+++ b/canister/src/api/get_utxos.rs
@@ -145,6 +145,17 @@ fn get_utxos_internal(
     }
 }
 
+// Returns the stability count of the given `target_block`.
+//
+// The stability count of a block is defined as the largest 𝜹 so that the block is 𝜹-stable.
+// A block b is 𝜹-stable if the following conditions hold:
+//   * d(b) ≥ 𝜹
+//   * ∀ b’ ∈ B \ {b}, h(b’) = h(b): d(b) - d(b’) ≥ 𝜹
+//
+// It follows from the above definition that the stability count is:
+// ```
+//    d(b) - max(d(b') ∀ b' ∈ B)
+// ```
 fn get_stability_count(
     blocks_with_depths_on_the_same_height: &[(&Block, u32)],
     target_block: BlockHash,
--- a/canister/src/api/get_utxos.rs
+++ b/canister/src/api/get_utxos.rs
@@ -145,12 +145,12 @@ fn get_utxos_internal(
     }
 }
 
-fn blocks_on_the_same_height_max_confirmation(
-    blocks: &[(Block, u32)],
+fn max_confirmations_count(
+    blocks_with_confirmations: &[(Block, u32)],
     avoid_block: BlockHash,
 ) -> u32 {
     let mut max_confirmations = 0;
-    for (block, confirmations) in blocks.iter() {
+    for (block, confirmations) in blocks_with_confirmations.iter() {
         if block.block_hash() != avoid_block {
             max_confirmations = std::cmp::max(max_confirmations, *confirmations);
         }
@@ -193,16 +193,14 @@ fn get_utxos_from_chain(
         let block_height = state.utxos.next_height() + (i as u32);
         let confirmations = chain_height - block_height + 1;
 
-        let max_confirmations_on_the_same_height = blocks_on_the_same_height_max_confirmation(
-            &blocks_with_confirmations_by_height[i],
-            block.block_hash(),
-        );
+        let max_confirmations_on_the_same_height =
+            max_confirmations_count(&blocks_with_confirmations_by_height[i], block.block_hash());
 
         let stability_count = confirmations - max_confirmations_on_the_same_height;
 
         if stability_count < min_confirmations {
-            // The block has fewer confirmations than requested.
-            // We can stop now since all remaining blocks will have fewer confirmations.
+            // The block has the lower stability count than requested.
+            // We can stop now since all remaining blocks will have lower stability count.
             break;
         }
 
--- a/canister/src/api/get_utxos.rs
+++ b/canister/src/api/get_utxos.rs
@@ -1221,10 +1221,10 @@ mod test {
     #[test]
     fn test_get_stability_count_single_block_on_height() {
         let block = BlockBuilder::genesis().build();
-        let blocks_with_confirmations: Vec<(&Block, u32)> = vec![(&block, 1)];
+        let blocks_with_depths: Vec<(&Block, u32)> = vec![(&block, 1)];
         // Stability count should be 1.
         assert_eq!(
-            get_stability_count(&blocks_with_confirmations, block.block_hash()),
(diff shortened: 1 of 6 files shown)


canister/src/api/get_utxos.rs
```diff
@@ -627,6 +650,7 @@ mod test {
     }
 
     #[test]
+    #[ignore]
```
dragoljub-duric, line 653: @ielashi Is this test obsolete now?

canister/src/api/get_utxos.rs
```diff
@@ -145,6 +145,19 @@ fn get_utxos_internal(
     }
 }
 
+fn max_confirmations_count(
```
ielashi, line 148: * I'd replace the term confirmations here with "depth". That's the more accurate term, and that is also the term used in the wiki.
* I think it'd be a better abstraction if we have a method called `get_stability_count` that encapsulates everything that's needed to compute the stability count, rather than having part of the logic in this method and the other part in `get_utxos_from_chain`.

canister/src/unstable_blocks.rs
```diff
@@ -71,6 +71,19 @@ impl UnstableBlocks {
         self.network
     }
 
+    /// Returns all blocks in the tree with the number of confirmations
+    /// separated by heights.
+    pub fn blocks_with_confirmations_by_height(
+        &self,
+        track_until_height: usize,
```
ielashi, line 78: Do we need this `track_until_height` parameter? In practice, it's almost always the case that the main chain is the longest chain, so we'd need to go through the entire `BlockTree`. I'd suggest removing this parameter to simplify the method.

canister/src/unstable_blocks.rs
```diff
@@ -71,6 +71,19 @@ impl UnstableBlocks {
...
+    ) -> Vec<Vec<(&Block, u32)>> {
+        let mut blocks_with_confirmations_by_height: Vec<Vec<(&Block, u32)>> =
+            vec![vec![]; track_until_height];
+        self.tree
+            .blocks_with_depth_by_height(&mut blocks_with_confirmations_by_height, 0);
+        blocks_with_confirmations_by_height
```
ielashi, line 84: I'd move this logic into the `BlockTree`, as that is code required to use `blocks_with_depth_by_height`. Adding unit tests to it would also be nice :)

canister/src/api/get_utxos.rs
ielashi: I think it's still useful to keep as this test is also focusing a lot on the `get_utxos` API, which isn't covered as well in `confirmation_counts.rs`.

canister/src/api/get_utxos.rs
dragoljub-duric: That is a good idea, I will do that.

canister/src/unstable_blocks.rs
dragoljub-duric: You are right, will do it.

canister/src/api/get_utxos.rs
dragoljub-duric: Are you sure that we should use "depth" everywhere, instead of "confirmation"? Because by the article in [wiki](https://wiki.internetcomputer.org/wiki/Bitcoin_integration), the depth of the tree with only the root node is 1, which is opposite to the general definition of the tree depth (in this case 0). Do you think it may be confusing to the reader? What do you think about using "confirmation" in code as well as in the wiki instead of "depth"?

canister/src/unstable_blocks.rs
dragoljub-duric: Done :)

canister/src/blocktree.rs
```diff
@@ -384,4 +415,118 @@ mod test {
...
+        assert_eq!(depth, 1);
+    }
+
+    #[test]
+    fn test_blocks_with_depths_separated_by_heights_chain() {
+        let mut blocks = vec![BlockBuilder::genesis().build()];
+        let chain_len = 10;
+        for i in 1..chain_len {
+            blocks.push(BlockBuilder::with_prev_header(blocks[i - 1].header()).build())
+        }
```
ielashi, line 443: You may find it simpler to use the `BlockchainBuilder`:

```suggestion
        let chain_len = 10;
        let blocks = let chain = BlockChainBuilder::new(chain_len).build();
```

canister/src/blocktree.rs
```diff
@@ -384,4 +415,118 @@ mod test {
...
+        let mut block_tree = BlockTree::new(genesis_block.clone());
+
+        let fork_1 = BlockBuilder::with_prev_header(genesis_block.header()).build();
+        extend(&mut block_tree, fork_1.clone()).unwrap();
+
+        let fork_2 = BlockBuilder::with_prev_header(genesis_block.header()).build();
+        extend(&mut block_tree, fork_2.clone()).unwrap();
+
+        let fork_2_extend = BlockBuilder::with_prev_header(fork_2.header()).build();
+        extend(&mut block_tree, fork_2_extend.clone()).unwrap();
```
ielashi, line 487: ditto (untested)

```suggestion
        let chain = BlockChainBuilder::new(2).build();
        // Create a fork from the genesis block with length 2.
        let fork = BlockChainBuilder::fork(&chain[0], 2).build();

```

canister/src/api/get_utxos.rs
```diff
@@ -1192,4 +1210,40 @@ mod test {
         // Base fee + instructions are charged for.
         assert_eq!(runtime::get_cycles_balance(), 10 + 1000);
     }
+
+    #[test]
+    fn test_get_stability_count_single_block_on_height() {
+        let block = BlockBuilder::genesis().build();
+        let blocks_with_confirmations: Vec<(&Block, u32)> = vec![(&block, 1)];
```
ielashi, line 1217:
```suggestion
        let blocks_with_depths: Vec<(&Block, u32)> = vec![(&block, 1)];
```
(ditto elsewhere in other tests)

canister/src/blocktree.rs
```diff
@@ -88,6 +88,37 @@ impl BlockTree {
             children: vec![],
         }
     }
+
+    /// Returns all blocks in the tree with their depths
+    /// separated by heights.
+    pub fn blocks_with_depths_separated_by_heights(&self) -> Vec<Vec<(&Block, u32)>> {
+        let mut blocks_with_confirmations_by_height: Vec<Vec<(&Block, u32)>> = vec![vec![]];
```
ielashi, line 95:
```suggestion
        let mut blocks_with_depths_by_height: Vec<Vec<(&Block, u32)>> = vec![vec![]];
```

Let's stick with depth rather that confirmations. I agree it's not exactly the same as the typical definition of depth in a CS tree, but bitcoin uses different definitions of height/depth anyways, so we can be consistent with those definitions.

ielashi (commented):
LGTM. Can we please also run `bash e2e-tests/profile.sh` and commit the changes in this PR?

PR #141, opened 2023-01-23
chore: make the network field in `UnstableBlocks` non-optional

dragoljub-duric:
After the latest release, the `network` field in `UnstableBlocks` no longer needs to be optional.

Closes [EXC-1310](https://dfinity.atlassian.net/browse/EXC-1310).


ielashi:
LGTM. Please don't forget to use [conventional commits](https://www.conventionalcommits.org/) in PR titles :)

ielashi (approved):

PR #144, opened 2023-01-26
chore: Track cycles received by the canister in a metric

dragoljub-duric:


canister/src/runtime.rs
```diff
@@ -190,8 +190,6 @@ pub fn msg_cycles_accept(max_amount: u64) -> u64 {
     max_amount
 }
 
-#[cfg(test)]
-#[cfg(not(target_arch = "wasm32"))]
 pub fn get_cycles_balance() -> u64 {
     CYCLES_BALANCE.with(|c| *c.borrow())
```
ielashi, line 194: This `CYCLES_BALANCE` is a mock cycles balance that's used for testing, so that's not what you want to use. You'd want to use https://docs.rs/ic-cdk/latest/ic_cdk/api/fn.canister_balance.html instead.

dragoljub-duric: Thank you. It is changed.

ielashi (approved):

PR #145, opened 2023-01-26
fix: Replace traps with rejects to correctly charge for cycles

dragoljub-duric:
Closes [EXC-1326].

Problem:
We did not charge cycles in the case of errors.

Solution:
Rather than trapping, we should instead explicitly reject so that we retain the cycles received in the request.

### DIFF
--- a/canister/src/api/get_balance.rs
+++ b/canister/src/api/get_balance.rs
@@ -153,9 +153,8 @@ mod test {
                 get_balance(GetBalanceRequest {
                     address: address.to_string(),
                     min_confirmations: *min_confirmations
-                })
-                .unwrap(),
-                1000
+                }),
+                Ok(1000)
             );
         }
 
@@ -193,7 +192,7 @@ mod test {
         }
 
         // The chain only contains the genesis block, so a min_confirmations of 2
-        // should return error, as there aren't that many blocks in the chain.
+        // should return an error, as there aren't that many blocks in the chain.
         assert_eq!(
             get_balance(GetBalanceRequest {
                 address: address.to_string(),
--- a/canister/src/api/get_balance.rs
+++ b/canister/src/api/get_balance.rs
@@ -18,11 +18,11 @@ struct Stats {
 }
 
 /// Retrieves the balance of the given Bitcoin address.
-pub fn get_balance(request: GetBalanceRequest) -> Satoshi {
+pub fn get_balance(request: GetBalanceRequest) -> Result<Satoshi, GetBalanceError> {
     verify_has_enough_cycles(with_state(|s| s.fees.get_balance_maximum));
     charge_cycles(with_state(|s| s.fees.get_balance));
 
-    get_balance_internal(request).expect("get_balance failed")
+    get_balance_internal(request)
 }
 
 fn get_balance_internal(request: GetBalanceRequest) -> Result<Satoshi, GetBalanceError> {
@@ -108,18 +108,20 @@ mod test {
     };
 
     #[test]
-    #[should_panic(expected = "get_balance failed: MalformedAddress")]
-    fn panics_on_malformed_address() {
+    fn error_on_malformed_address() {
         crate::init(Config {
             stability_threshold: 1,
             network: Network::Mainnet,
             ..Default::default()
         });
 
-        get_balance(GetBalanceRequest {
-            address: String::from("not an address"),
-            min_confirmations: None,
-        });
+        assert_eq!(
+            get_balance(GetBalanceRequest {
+                address: String::from("not an address"),
+                min_confirmations: None,
+            }),
+            Err(GetBalanceError::MalformedAddress)
+        );
     }
 
     #[test]
@@ -151,7 +153,8 @@ mod test {
                 get_balance(GetBalanceRequest {
                     address: address.to_string(),
                     min_confirmations: *min_confirmations
-                }),
+                })
+                .unwrap(),
                 1000
             );
         }
@@ -161,14 +164,14 @@ mod test {
             get_balance(GetBalanceRequest {
                 address: address.to_string(),
                 min_confirmations: Some(2)
-            }),
+            })
+            .unwrap(),
             0
         );
     }
 
     #[test]
-    #[should_panic(expected = "get_balance failed: MinConfirmationsTooLarge { given: 2, max: 1 }")]
-    fn panics_on_very_large_confirmations() {
+    fn error_on_very_large_confirmations() {
         let network = Network::Regtest;
         crate::init(Config {
             stability_threshold: 2,
@@ -183,17 +186,21 @@ mod test {
                 get_balance(GetBalanceRequest {
                     address: address.to_string(),
                     min_confirmations
-                }),
+                })
+                .unwrap(),
                 0
             );
         }
 
         // The chain only contains the genesis block, so a min_confirmations of 2
-        // should panic, as there aren't that many blocks in the chain.
-        get_balance(GetBalanceRequest {
-            address: address.to_string(),
-            min_confirmations: Some(2),
-        });
+        // should return error, as there aren't that many blocks in the chain.
+        assert_eq!(
+            get_balance(GetBalanceRequest {
+                address: address.to_string(),
+                min_confirmations: Some(2),
+            }),
+            Err(GetBalanceError::MinConfirmationsTooLarge { given: 2, max: 1 })
+        );
     }
 
     #[test]
@@ -241,7 +248,8 @@ mod test {
                 get_balance(GetBalanceRequest {
                     address: address_2.to_string(),
                     min_confirmations: *min_confirmations
-                }),
+                })
+                .unwrap(),
                 1000
             );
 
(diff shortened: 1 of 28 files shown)


dragoljub-duric:
Note:

1. I am not sure whether it is possible to test when the canister rejects [here](https://github.com/dfinity/bitcoin-canister/blob/master/canister/src/api/send_transaction.rs#L37).
2. Fees used in the test are not standard ones, hence cycle charging is not realistic.

canister/src/api/get_balance.rs
```diff
@@ -183,17 +186,21 @@ mod test {
...
             );
         }
 
         // The chain only contains the genesis block, so a min_confirmations of 2
-        // should panic, as there aren't that many blocks in the chain.
-        get_balance(GetBalanceRequest {
-            address: address.to_string(),
-            min_confirmations: Some(2),
-        });
+        // should return error, as there aren't that many blocks in the chain.
```
ielashi, line 196:
```suggestion
        // should return an error, as there aren't that many blocks in the chain.
```

canister/src/api/get_utxos.rs
```diff
@@ -644,11 +658,14 @@ mod test {
         }
 
         // The chain only contains the genesis block, so a min_confirmations of 2
-        // should panic, as there aren't that many blocks in the chain.
-        get_utxos(GetUtxosRequest {
-            address: address.to_string(),
-            filter: Some(UtxosFilter::MinConfirmations(2)),
-        });
+        // should return error, as there aren't that many blocks in the chain.
```
ielashi, line 661:
```suggestion
        // should return an error, as there aren't that many blocks in the chain.
```

canister/src/api/get_utxos.rs
```diff
@@ -930,11 +958,14 @@ mod test {
             }
         );
 
-        // min confirmations is too large. Should panic.
-        get_utxos(GetUtxosRequest {
-            address: address_1.to_string(),
-            filter: Some(UtxosFilter::MinConfirmations(3)),
-        });
+        // min confirmations is too large. Should return error.
```
ielashi, line 961:
```suggestion
        // min confirmations is too large. Should return an error.
```

canister/src/api/send_transaction.rs
```diff
@@ -34,7 +34,12 @@ pub async fn send_transaction(request: SendTransactionRequest) {
         },
     )
     .await
-    .expect("Sending transaction bitcoin network must succeed");
+    .map_err(|(code, msg)| {
+        format!(
+            "Sending transaction bitcoin network must succeed, Rejection code: {:?}, {}",
+            code, msg
+        )
+    })
```
ielashi, line 42: This is not an error based on user input, so the correct behavior here would be to trap and not reject, so I'd revert the changes here.

canister/src/api/get_balance.rs
```diff
@@ -151,7 +153,8 @@ mod test {
                 get_balance(GetBalanceRequest {
                     address: address.to_string(),
                     min_confirmations: *min_confirmations
-                }),
+                })
+                .unwrap(),
                 1000
```
ielashi, line 158: nit: instead of unwrapping, I suggest we compare with `Ok(_)`, which returns an easier to read error if the assertion fails. Ditto everywhere else.

```suggestion
                }),
                Ok(1000)
```

canister/src/api/get_utxos.rs
```diff
@@ -114,6 +113,15 @@ fn get_utxos_internal(
...
         Some(page) => {
+            if page.len() != 72 {
+                return Err(GetUtxosError::MalformedPage {
+                    err: format!(
+                        "Could not parse page, page size {}, expected size {}.",
+                        page.len(),
+                        72
+                    ),
+                });
+            }
```
ielashi, line 124: I'd move this logic inside `Page::from_bytes` as that's where it logically belongs. I'd also define a constant for the expected page length and use that (i.e. `const EXPECTED_PAGE_LENGTH: usize = 72`.

canister/src/api/send_transaction.rs
```diff
@@ -5,7 +5,7 @@ use crate::{
 use bitcoin::{consensus::Decodable, Transaction};
 use ic_btc_types::SendTransactionRequest;
 
-pub async fn send_transaction(request: SendTransactionRequest) {
+pub async fn send_transaction(request: SendTransactionRequest) -> Result<(), String> {
```
ielashi, line 8: When possible, we should not return errors as strings, as these are hard to parse, and the caller can't easily distinguish between the various cases. I know in this case there's only one error, so it doesn't matter too much, but it's good to stay principled and consistent. I'd define an enum for the error, and that translate that error into a string in the calling function.

```
enum SendTransactionError {
  CannotDecode
}
```

e2e-tests/charge-cycles-on-reject.sh
```diff
@@ -0,0 +1,102 @@
...
+    exit 1
+  fi
+
+  AFTER_SEND_TRANSACTION=$(dfx wallet balance --precise)
+
+  # Should charge cycles.
+  if [[ $BEFORE_SEND_TRANSACTION = "$AFTER_SEND_TRANSACTION" ]]; then
+    echo "FAIL"
+    exit 1
+  fi
```
ielashi, line 62: I'd check that the difference in cycles is exactly the fee we're expecting to make the test more rigorous. And, related to this, I'd simply use a fee of `1` for everything when initializing the canister to make the math here simple.

e2e-tests/charge-cycles-on-reject.sh
```diff
@@ -0,0 +1,102 @@
...
+
+# Deploy the canister that returns the blocks for harge-cycles-on-reject.
+dfx deploy e2e-scenario-3
+
+# Deploy the bitcoin canister, setting the blocks_source to be the source above.
+dfx deploy bitcoin --argument "(record {
+  stability_threshold = 2;
+  network = variant { regtest };
+  blocks_source = principal \"$(dfx canister id e2e-scenario-3)\";
+  syncing = variant { enabled };
```
ielashi, line 20: nit: We don't need to be syncing blocks in this test.

```suggestion
  syncing = variant { disabled };
```

canister/src/api/get_utxos.rs
dragoljub-duric: It looks like I will finally need to learn articles:)

canister/src/api/send_transaction.rs
dragoljub-duric: Good point, I agree.

canister/src/api/send_transaction.rs
dragoljub-duric: Done.

e2e-tests/charge-cycles-on-reject.sh
dragoljub-duric: Done, please take a look now.

canister/src/api/get_utxos.rs
```diff
@@ -35,9 +35,12 @@ struct Stats {
 }
 
 /// Retrieves the UTXOs of the given Bitcoin address.
-pub fn get_utxos(request: GetUtxosRequest) -> GetUtxosResponse {
+pub fn get_utxos(request: GetUtxosRequest) -> Result<GetUtxosResponse, GetUtxosError> {
     verify_has_enough_cycles(with_state(|s| s.fees.get_utxos_maximum));
 
+    // Charge the base fee.
+    charge_cycles(with_state(|s| s.fees.get_utxos_base));
```
dragoljub-duric, line 42: @ielashi I needed to make those changes to charge some cycles when rejected. But since we do not have more information, we can only charge a base fee and charge for the instructions only when the transaction succeeds.

e2e-tests/charge-cycles-on-reject.sh
```diff
@@ -0,0 +1,106 @@
...
+
+SCRIPT_DIR="$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
+source "${SCRIPT_DIR}/utils.sh"
+
+# Run dfx stop if we run into errors.
+trap "dfx stop" EXIT SIGINT
+
+dfx start --background --clean
+
+# Deploy the canister that returns the blocks for charge-cycles-on-reject.
```
ielashi, line 12: I'm not sure what this means. We are not fetching blocks from this canister at all, right? Isn't the reason here that we want to send the `send_transaction` endpoint? If so, can you clarify that in the comment?

ielashi (approved):
LGTM.

Please note my comment on the PR title/description from the above PR:

> LGTM. A couple of comments (and this applies to other PRs as well):
> 
> 1. Use [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/).
> 2. Do not add JIRA links in the description. You can write the ticket number, but do not link to it. External viewers don't have access to these links and it's an unnecessary endorsement of JIRA. Instead, please write a description of the intention behind the commit. Personally, I like to write commits in the following format, as I find that it helps the reader get context on what the change is and why we're doing it.
> 
> ```
> # Problem
> ...
> 
> # Solution
> ...
> ```

e2e-tests/charge-cycles-on-reject.sh
dragoljub-duric: You are right, I have changed that.

PR #147, opened 2023-02-03
fix: Set syncing flag in init method.

dragoljub-duric:
Closes EXC-1322.

Problem:
The init method currently ignores the syncing flag. When the canister starts, its syncing flag will always be enabled even if syncing is set to disabled in the init parameters.

Solution:
To fix this, the init method needs to update the syncing flag based on the given config.

### DIFF
--- a/canister/src/lib.rs
+++ b/canister/src/lib.rs
@@ -83,6 +83,7 @@ pub fn init(config: Config) {
 
     with_state_mut(|s| s.blocks_source = config.blocks_source);
     with_state_mut(|s| s.api_access = config.api_access);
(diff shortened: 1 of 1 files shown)


ielashi (approved):
LGTM. A couple of comments (and this applies to other PRs as well):

1. Use [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/).
2. Do not add JIRA links in the description. You can write the ticket number, but do not link to it. External viewers don't have access to these links and it's an unnecessary endorsement of JIRA. Instead, please write a description of the intention behind the commit. Personally, I like to write commits in the following format, as I find that it helps the reader get context on what the change is and why we're doing it.

```
# Problem
...

# Solution
...
```

PR #149, opened 2023-02-07
fix: Validate that the block is < 2 hours from the current time.

dragoljub-duric:
Problem:
We’ll need to update [this logic](https://github.com/dfinity/bitcoin-canister/blob/master/validation/src/header.rs#L92) to check if the timestamp of the block is < 2 hours from the current time.

Solution:
This logic is shared between both the canister and the adapter, and we’d want this code to work in both environments. For the canister environment (wasm32), we can use [time in ic_cdk::api - Rust](https://docs.rs/ic-cdk/latest/ic_cdk/api/fn.time.html)  to get the current time, for other environments we can use [SystemTime in std::time - Rust](https://doc.rust-lang.org/std/time/struct.SystemTime.html).


validation/src/header.rs
```diff
@@ -86,10 +87,44 @@ pub fn validate_header(
...
+    let number_of_seconds_in_one_hour = 60 * 60;
+    let time_diff_hours = time_diff_secs / number_of_seconds_in_one_hour;
+    time_diff_hours < 2
+}
+
 /// Validates if a header's timestamp is valid.
 /// Bitcoin Protocol Rules wiki https://en.bitcoin.it/wiki/Protocol_rules says,
 /// "Reject if timestamp is the median time of the last 11 blocks or before"
-fn is_timestamp_valid(store: &impl HeaderStore, header: &BlockHeader) -> bool {
+/// "Reject if the timestamp of the block is < 2 hours from the current time"
```
ielashi, line 120: I'd quote what the page above states exactly: "Block timestamp must not be more than two hours in the future".

validation/src/header.rs
```diff
@@ -441,14 +515,40 @@ mod test {
             bits: 0x170e0408,
             nonce: 0xb48e8b0a,
         };
-        assert!(is_timestamp_valid(&store, &header));
+        assert!(matches!(is_timestamp_valid(&store, &header), Ok(())));
```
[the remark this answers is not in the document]
dragoljub-duric: That does not compile:
`error[E0369]: binary operation `==` cannot be applied to type `Result<(), header::ValidateHeaderError>`
   --> validation/src/header.rs:518:9
    |
518 |         assert_eq!(is_timestamp_valid(&store, &header), Ok(()));
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |         |
    |         Result<(), header::ValidateHeaderError>
    |         Result<(), header::ValidateHeaderError>`

validation/Cargo.toml
```diff
@@ -7,6 +7,7 @@ edition = "2021"
 
 [dependencies]
 bitcoin = "0.28.1"
+ic-cdk = "0.6.1"
```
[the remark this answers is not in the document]
dragoljub-duric: Good point, I will change that.

validation/src/header.rs
```diff
@@ -13,6 +13,9 @@ pub enum ValidateHeaderError {
     /// Used when the timestamp in the header is lower than
     /// the median of timestamps of past 11 headers.
     HeaderIsOld,
+    /// Used when the timestamp in the header is more than 2 hours
+    /// from the current time.
+    HeaderIsMoreThan2HFromCurrentTime,
```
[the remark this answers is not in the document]
dragoljub-duric: Yes, I can do that. But I will change the name from `HeaderIsInFuture` to `HeaderIsTooFarInFuture` because it can be in the future and still pass the check, hence name could be misleading. :)

validation/Cargo.toml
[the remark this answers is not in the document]
dragoljub-duric: Done.

validation/src/header.rs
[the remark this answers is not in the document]
dragoljub-duric: maybe to use `assert!(is_timestamp_valid(&store, &header).is_ok())`

validation/src/header.rs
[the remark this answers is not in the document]
dragoljub-duric: Done.

canister/src/state.rs
```diff
@@ -90,6 +90,22 @@ impl State {
...
+
+#[cfg(not(target_arch = "wasm32"))]
+fn get_current_time_in_secs() -> u64 {
+    use std::time::SystemTime;
+
+    SystemTime::now()
+        .duration_since(SystemTime::UNIX_EPOCH)
+        .unwrap()
+        .as_secs()
+}
```
ielashi, line 107: Is `state.rs` the right place to put this logic? I'd create a method called `fn time()` in `runtime.rs` that returns the time in ns. That's where we put logic that is different across runtimes.

validation/src/constants.rs
```diff
@@ -84,4 +84,6 @@ pub mod test {
     pub const TESTNET_HEADER_2132555: &str = "004000200e1ff99438666c67c649def743fb82117537c2017bcc6ad617000000000000007fa40cf82bf224909e3174281a57af2eb3a4a2a961d33f50ec0772c1221c9e61ddfdc061ffff001a64526636";
     /// Testnet 00000000383cd7fff4692410ccd9bd6201790043bb41b93bacb21e9b85620767
     pub const TESTNET_HEADER_2132556: &str = "00000020974f55e77dff100bc252a01aa7b00d16736c6e04a091b03be200000000000000c44f2d69fc200c4a2211885000b6b67512f42c1bec550f3754e103b6c4046e05a202c161ffff001d09ec1bc4";
+    /// Mock time used as the current time.
+    pub const MOCK_CURRENT_TIME: u64 = 2_634_590_600;
```
ielashi, line 88: Why is this constant here? It's not used in this file, and it's also not used in production. I'd keep this constant with the tests where it's used.

validation/src/header.rs
```diff
@@ -441,22 +517,45 @@ mod test {
...
+        let one_hour = 60 * 60;
+
+        header.time = (curr_time - one_hour) as u32;
+
+        assert!(is_timestamp_valid(&store, &header, curr_time).is_ok());
+
+        header.time = (curr_time + 2 * one_hour + 10) as u32;
+        assert!(matches!(
+            is_timestamp_valid(&store, &header, curr_time),
+            Err(ValidateHeaderError::HeaderIsTooFarInFuture { .. })
```
ielashi, line 543: nit: I'd expand these variables, as it makes the test stronger.

validation/src/header.rs
```diff
@@ -441,22 +517,45 @@ mod test {
...
+        ));
+
+        let result = validate_header(&Network::Bitcoin, &store, &header, curr_time);
+        assert!(matches!(
+            result,
+            Err(ValidateHeaderError::HeaderIsTooFarInFuture { .. })
```
ielashi, line 549: ditto

validation/src/header.rs
```diff
@@ -86,10 +91,33 @@ pub fn validate_header(
     Ok(())
 }
 
+fn block_timestamp_less_than_2h_from_current_time(
+    block_time: u64,
+    current_time: u64,
+) -> Result<(), ValidateHeaderError> {
+    let number_of_seconds_in_one_hour = 60 * 60;
+    let max_allowed_time = current_time + 2 * number_of_seconds_in_one_hour;
```
ielashi, line 99: please define the numbers here as a constant.

validation/src/header.rs
```diff
@@ -86,10 +91,33 @@ pub fn validate_header(
...
 /// Bitcoin Protocol Rules wiki https://en.bitcoin.it/wiki/Protocol_rules says,
 /// "Reject if timestamp is the median time of the last 11 blocks or before"
-fn is_timestamp_valid(store: &impl HeaderStore, header: &BlockHeader) -> bool {
+/// "Block timestamp must not be more than two hours in the future"
+fn is_timestamp_valid(
+    store: &impl HeaderStore,
+    header: &BlockHeader,
+    current_time: u64,
+) -> Result<(), ValidateHeaderError> {
+    block_timestamp_less_than_2h_from_current_time(header.time as u64, current_time)?;
```
ielashi, line 120: I think this method name is misleading. We're not verifying that the block timestamp is less than 2h from the current time.

validation/src/constants.rs
dragoljub-duric: Those are constants for `mod test`,  and all constants from above are used only in tests in `header.rs`.

canister/src/state.rs
dragoljub-duric: Done.

validation/src/header.rs
dragoljub-duric: Changed.

ielashi (approved):
LGTM. Please rerun the profiling script before merging.

PR #150, opened 2023-02-08
feat: Add metric to track block ingestion instruction counts

dragoljub-duric:
Problem:
We currently do not have a way of tracking how many instructions it takes to ingest new Bitcoin blocks.

Solution:
Keep track of the latest stats in a variable and expose them as a metric so that we can track these on our dashboards.

### DIFF
--- a/canister/src/metrics.rs
+++ b/canister/src/metrics.rs
@@ -24,7 +24,7 @@ pub struct Metrics {
     pub send_transaction_count: u64,
 
     // The number of instructions spent ingesting new Bitcoin blocks by type.
-    pub ingest_block_instruction_count: LabeledCounter,
+    pub ingest_block_instructions_count: LabeledCounter,
 }
 
 impl Default for Metrics {
@@ -59,8 +59,8 @@ impl Default for Metrics {
 
             send_transaction_count: 0,
 
-            ingest_block_instruction_count: LabeledCounter::new(
-                "ingest_block_instruction_count",
+            ingest_block_instructions_count: LabeledCounter::new(
+                "ingest_block_instructions_count",
                 "The number of instructions spent ingesting new Bitcoin blocks by type.",
                 BlockIngestionStats::get_all_labels(),
             ),
--- a/canister/src/metrics.rs
+++ b/canister/src/metrics.rs
@@ -1,4 +1,8 @@
+use std::collections::HashMap;
+
 use serde::{Deserialize, Serialize};
+
+use crate::utxo_set::BlockIngestionStats;
 
 const M: u64 = 1_000_000;
 const BUCKET_SIZE: u64 = 500 * M;
@@ -19,8 +23,8 @@ pub struct Metrics {
     /// The total number of (valid) requests sent to `send_transaction`.
     pub send_transaction_count: u64,
 
-    // The total number of instructions spent ingesting new Bitcoin blocks.
-    pub total_block_ingestion_instruction_count: u64,
+    // The number of instructions spent ingesting new Bitcoin blocks by type.
+    pub ingest_block_instruction_count: LabeledCounter,
 }
 
 impl Default for Metrics {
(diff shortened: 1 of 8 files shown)


canister/src/utxo_set.rs
```diff
@@ -512,6 +515,28 @@ struct BlockIngestionStats {
...
+            String::from("ins_remove_inputs"),
+            String::from("ins_insert_outputs"),
+            String::from("ins_txids"),
+            String::from("ins_insert_utxos"),
+        ]
+    }
+
+    pub fn get_labels_and_values(&self) -> Vec<(String, u64)> {
+        vec![
+            (String::from("ins_total"), self.ins_total),
```
dragoljub-duric, line 531: @ielashi I think `ins_total` is unnecessary because I guess Prometheus will also calculate the sum of all types. What do you think?

ielashi (commented):
Thanks @dragoljub-duric! Question: how did we test this? As in, how did we ensure that whatever gets output in the `/metrics` endpoint is valid and can be read correctly by prometheus?

canister/src/utxo_set.rs
```diff
@@ -602,11 +628,19 @@ mod test {
...
+                    assert_eq!(hash, block.block_hash());
+                    // Cost of inserting one output (1000)
+                    assert_eq!(stats.ins_total, 1000);
+                    assert_eq!(stats.ins_insert_outputs, 1000);
+                    assert_eq!(stats.ins_insert_utxos, 0);
+                    assert_eq!(stats.ins_remove_inputs, 0);
+                    assert_eq!(stats.ins_txids, 0);
+                }
+                _ => panic!("Unexpected result."),
+            }
```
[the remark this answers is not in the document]
dragoljub-duric: Good point, done.

canister/src/metrics.rs
```diff
@@ -18,6 +22,9 @@ pub struct Metrics {
 
     /// The total number of (valid) requests sent to `send_transaction`.
     pub send_transaction_count: u64,
+
+    // The number of instructions spent ingesting new Bitcoin blocks by type.
+    pub ingest_block_instructions_count: LabeledCounter,
```
ielashi, line 27:
```suggestion
    // The number of instructions spent ingesting the most recent stable block.
    pub ingest_block_instructions_count: LabeledCounter,
```

dragoljub-duric: Are you sure about this comment, 'the most recent stable block' is just one block, while this counter will be the sum of instructions used for all ingested blocks?

canister/src/state.rs
```diff
@@ -267,11 +280,35 @@ mod test {
...
+                // no block as ingested, hence the number of instruction should be 0
+                0
+            });
+
+            // We have only output transactions, hence 'ins_insert_outputs' should be equal to 'ins_total'.
+            assert_eq!(ins_total_after - ins_total_before, ins_insert_outputs_after - ins_insert_outputs_before);
+            // While all others should be 0.
+            assert_eq!(*state.metrics.ingest_block_instructions_count.get_value_with_label("ins_remove_inputs"), 0);
+            assert_eq!(*state.metrics.ingest_block_instructions_count.get_value_with_label("ins_txids"), 0);
+            assert_eq!(*state.metrics.ingest_block_instructions_count.get_value_with_label("ins_insert_utxos"), 0);
```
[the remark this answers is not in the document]
dragoljub-duric: I move it to a separate test.

canister/src/metrics.rs
ielashi: Oh I missed the part where we were summing all the instruction counts. I thought we were only keeping track of the most recent block ingestion, which is what would be more useful for our dashboards. I went ahead and made the following changes:

1. We're now only recording the most recent block ingestion (i.e. not summing)
2. Using a gauge instead of a counter (which is necessary since the value of the instruction counts can go up and down).
3. I updated the code to use the `ic_metrics_encoder` crate while we were at it - mostly simple changes.
4. I removed the proptest in `state.rs`, since now that we aren't summing the blocks anymore the test didn't contribute much.

ielashi (approved):

ielashi (approved):

canister/src/metrics.rs
dragoljub-duric: Regarding the first point: TBH, I am not convinced that recording the most recent block ingestion is better than summing, for example:

  1. one ingestion with I = (i1, i2, i3, ...) instructions used
  2. n ingestions with I = (i1, i2, i3, ...) instructions used
  
We will not see any difference on our Prometheus dashboards, and AFAIK it will not be possible to get the cumulative of instructions per type since we do not have any field to sum based on it.

Regarding the last point, that test was intended to verify that metrics are correctly updated after executing `ingest_block`, without that test if we delete the places on which we change the value of the metric, all tests will pass, but we do not have the intended behavior.

dragoljub-duric: Regarding 3rd point, thanks for changing that.

ielashi: @dragoljub-duric As discussed offline, I kept the gauge and re-added the test in `state.rs`. I modified it a bit so that it's testing specifically the two cases where we update the stats.

PR #151, opened 2023-03-13
feat: Do not respond to requests when not fully synced

dragoljub-duric:
Problem:

Currently, the Bitcoin canister serves requests even if it is significantly behind the tip, making it risky that dapps would base their decisions on outdated information.

Solution:

The Bitcoin canister is in the state “synced” if the difference between the maximum height of all block headers and the maximum height of all unstable blocks is at most 2. Otherwise, the canister is in the state “syncing”.

The Bitcoin canister only acts on requests if it is in the state “synced”, and otherwise traps if it isn’t synced.

### DIFF
--- a/canister/src/heartbeat.rs
+++ b/canister/src/heartbeat.rs
@@ -32,12 +32,16 @@ pub async fn heartbeat() {
 // Fetches new blocks if there isn't a request in progress and no complete response to process.
 // Returns true if a call to the `blocks_source` has been made, false otherwise.
 async fn maybe_fetch_blocks() -> bool {
-    if with_state(|s| {
-        s.syncing_state.is_fetching_blocks || s.syncing_state.syncing == Flag::Disabled
-    }) {
-        // Already fetching blocks or syncing is disabled.
+    if with_state(|s| s.syncing_state.syncing == Flag::Disabled) {
+        // Syncing is disabled.
         return false;
     }
+
+    // A guard to verify we aren't already fetching blocks.
+    let _guard = match crate::guard::FetchBlocksGuard::new() {
+        Some(guard) => guard,
+        None => return false,
+    };
 
     // Request additional blocks.
     let maybe_request = maybe_get_successors_request();
@@ -49,11 +53,6 @@ async fn maybe_fetch_blocks() -> bool {
         }
     };
 
-    // A lock to ensure the heartbeat only sends one request at a time.
-    with_state_mut(|s| {
-        s.syncing_state.is_fetching_blocks = true;
-    });
-
     print(&format!("Sending request: {:?}", request));
 
     let response: Result<(GetSuccessorsResponse,), _> =
@@ -61,10 +60,8 @@ async fn maybe_fetch_blocks() -> bool {
 
     print(&format!("Received response: {:?}", response));
 
-    // Release the heartbeat lock and save the response.
+    // Save the response.
     with_state_mut(|s| {
-        s.syncing_state.is_fetching_blocks = false;
-
         let response = match response {
             Ok((response,)) => response,
             Err((code, msg)) => {
--- a/canister/src/heartbeat.rs
+++ b/canister/src/heartbeat.rs
@@ -165,18 +165,17 @@ fn maybe_process_response() {
                     }
                 }
                 for block_header_blob in response.next.iter() {
-                    let block_header = match BlockHeader::consensus_decode(
-                        block_header_blob.as_slice(),
-                    ) {
-                        Ok(header) => header,
-                        Err(err) => {
-                            print(&format!(
+                    let block_header =
+                        match BlockHeader::consensus_decode(block_header_blob.as_slice()) {
+                            Ok(header) => header,
+                            Err(err) => {
+                                print(&format!(
                                 "ERROR: Failed decode block header. Err: {:?}, Block header: {:?}",
                                 err, block_header_blob,
                             ));
-                            break;
-                        }
-                    };
+                                break;
+                            }
+                        };
 
                     let target = block_header.target();
                     if let Err(err) = block_header.validate_pow(&target) {
--- a/canister/src/heartbeat.rs
+++ b/canister/src/heartbeat.rs
@@ -156,7 +156,7 @@ fn maybe_process_response() {
                         }
                     };
 
-                    if let Err(err) = state::insert_block(state, Block::new(block)) {
+                    if let Err(err) = state::insert_block(state, Block::new(block.clone())) {
                         print(&format!(
                             "ERROR: Failed to insert block. Err: {:?}, Block bytes: {:?}",
                             err, block_bytes,
--- a/canister/src/heartbeat.rs
+++ b/canister/src/heartbeat.rs
@@ -1,6 +1,6 @@
 use crate::{
     runtime::{call_get_successors, print},
-    state::{self, insert_next_block, remove_received_block_from_next_blocks, ResponseToProcess},
+    state::{self, ResponseToProcess},
     types::{
         Block, BlockHash, Flag, GetSuccessorsCompleteResponse, GetSuccessorsRequest,
         GetSuccessorsRequestInitial, GetSuccessorsResponse,
@@ -166,18 +166,17 @@ fn maybe_process_response() {
                         state.syncing_state.num_insert_block_errors += 1;
                         return;
                     }
-
-                    remove_received_block_from_next_blocks(
-                        state,
-                        &BlockHash::from(block.block_hash()),
-                    );
                 }
                 for block_header_blob in response.next.iter() {
                     let block_header = BlockHeader::consensus_decode(block_header_blob.as_slice())
                         .expect("block header decoding must succeed");
                     let block_hash = BlockHash::from(block_header.block_hash());
                     let prev_hash = BlockHash::from(block_header.prev_blockhash);
-                    insert_next_block(state, &prev_hash, &block_hash);
+                    state.unstable_blocks.insert_next_block(
+                        &prev_hash,
+                        &block_hash,
+                        state.stable_height(),
+                    );
                 }
             }
             other => {
--- a/canister/src/heartbeat.rs
+++ b/canister/src/heartbeat.rs
@@ -74,7 +74,7 @@ async fn maybe_fetch_blocks() -> bool {
                 return;
             }
         };
-
+        //EXC-980
         match response {
             GetSuccessorsResponse::Complete(response) => {
                 // Received complete response.
@@ -176,6 +176,7 @@ fn maybe_process_response() {
     });
 }
 
+//EXC-980
 // Retrieves a `GetSuccessorsRequest` to send to the adapter.
 fn maybe_get_successors_request() -> Option<GetSuccessorsRequest> {
     with_state(|state| match &state.syncing_state.response_to_process {
--- a/canister/src/heartbeat.rs
+++ b/canister/src/heartbeat.rs
@@ -180,18 +180,24 @@ fn maybe_process_response() {
                     let target = block_header.target();
                     if let Err(err) = block_header.validate_pow(&target) {
                         print(&format!(
-                            "ERROR: Failed validate block header. Err: {:?}, Block header: {:?}",
+                            "ERROR: Failed to validate block header. Err: {:?}, Block header: {:?}",
                             err, block_header,
                         ));
                         break;
                     }
                     let block_hash = BlockHash::from(block_header.block_hash());
                     let prev_hash = BlockHash::from(block_header.prev_blockhash);
-                    state.unstable_blocks.insert_next_block_hash(
+                    if let Err(err) = state.unstable_blocks.insert_next_block_hash(
                         &prev_hash,
                         &block_hash,
                         state.stable_height(),
-                    );
+                    ) {
+                        print(&format!(
+                            "ERROR: Failed to insert next block hash. Err: {:?}, Block header: {:?}",
+                            err, block_header,
+                        ));
+                        break;
+                    }
                 }
             }
             other => {
--- a/canister/src/heartbeat.rs
+++ b/canister/src/heartbeat.rs
@@ -1,6 +1,6 @@
 use crate::{
     runtime::{call_get_successors, print},
-    state::{self, insert_expected_block, remove_received_expected_block, ResponseToProcess},
+    state::{self, insert_next_block, remove_received_block_from_next_blocks, ResponseToProcess},
     types::{
         Block, BlockHash, Flag, GetSuccessorsCompleteResponse, GetSuccessorsRequest,
         GetSuccessorsRequestInitial, GetSuccessorsResponse,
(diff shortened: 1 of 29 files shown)


canister/src/tests.rs
```diff
@@ -544,3 +544,191 @@ async fn test_rejections_counting() {
...
 }
+
+// Serialize header.
+fn get_header_blob(header: &BlockHeader) -> BlockHeaderBlob {
+    let mut header_buff = vec![];
+    header.consensus_encode(&mut header_buff).unwrap();
+    header_buff.into()
+}
+
+fn get_chain_with_n_block_and_header_blobs(
```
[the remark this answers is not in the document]
dragoljub-duric: Good point, done.

canister/src/heartbeat.rs
```diff
@@ -166,6 +166,15 @@ fn maybe_process_response() {
                         state.syncing_state.num_insert_block_errors += 1;
                         return;
                     }
+
+                    remove_received_expected_block(state, &BlockHash::from(block.block_hash()));
```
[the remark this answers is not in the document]
dragoljub-duric: Done.

canister/src/state.rs
```diff
@@ -136,6 +243,7 @@ pub fn ingest_stable_blocks_into_utxoset(state: &mut State) -> bool {
         Some(Slicing::Paused(())) => return has_state_changed(state),
         Some(Slicing::Done((ingested_block_hash, stats))) => {
             state.metrics.block_ingestion_stats = stats;
+            remove_expected_blocks_based_on_stable_height(state);
```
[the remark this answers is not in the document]
dragoljub-duric: Done.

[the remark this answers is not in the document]
dragoljub-duric: PTAL

canister/src/next_blocks.rs
```diff
@@ -0,0 +1,222 @@
...
+use std::collections::BTreeMap;
+
+#[derive(Serialize, Deserialize, Default, Debug, Clone, PartialEq)]
+pub struct NextBlocks {
+    hash_to_height: BTreeMap<BlockHash, Height>,
+    height_to_hash: BTreeMap<Height, Vec<BlockHash>>,
+}
+
+impl NextBlocks {
+    pub(crate) fn insert(&mut self, block_hash: &BlockHash, height: Height) {
```
ielashi, line 13: nit: you don't need the `(crate)` for these methods, because this whole module is private to the crate. Similarly for other methods here.

canister/src/unstable_blocks.rs
```diff
@@ -19,6 +20,8 @@ pub struct UnstableBlocks {
     tree: BlockTree,
     outpoints_cache: OutPointsCache,
     network: Network,
+    /// Blocks that are expected to be received.
```
ielashi, line 23:
```suggestion
    /// The hashes of the blocks that are expected to be received.
```

canister/src/state.rs
```diff
@@ -52,6 +54,8 @@ pub struct State {
     pub api_access: Flag,
 }
 
+pub const SYNCING_THRESHOLD: u32 = 2;
```
ielashi, line 57: docs. I'd also move this constant, along with the `is_fully_synced` into lib.rs, as it's not really part of the state.

canister/src/unstable_blocks.rs
```diff
@@ -71,6 +75,48 @@ impl UnstableBlocks {
...
+    }
+
+    // Removes blocks from 'next_blocks' with height up to current 'stable_height'.
+    pub fn remove_next_blocks_based_on_stable_height(&mut self, stable_height: Height) {
+        self.next_blocks.remove_until_height(stable_height);
+    }
+
+    // Public only for testing purpose.
+    pub fn next_blocks_max_height(&self) -> Height {
+        self.next_blocks.get_max_height().unwrap_or(0)
```
ielashi, line 118: I find it odd that we're returning a height of zero here when there are no next blocks. Wouldn't be more clear to the callers if this returned an `Option<Height>`.

canister/src/state.rs
```diff
@@ -89,6 +93,22 @@ impl State {
     pub fn get_utxos(&self, address: Address) -> AddressUtxoSet<'_> {
         AddressUtxoSet::new(address, &self.utxos, &self.unstable_blocks)
     }
+
+    /// Return 'true' if the difference between the maximum height of all block
+    /// headers and the maximum height of all unstable blocks is at most g.
```
ielashi, line 98: `g` is not defined here.

```suggestion
    /// Returns 'true' if the difference between the maximum height of all block
    /// headers and the maximum height of all unstable blocks is at most the syncing threshold.
```

ielashi (commented):
Thanks @dragoljub-duric. I'm not fully done with the review yet. I left some (mostly minor) comments and will give it another pass once you've taken a look at them.

canister/src/heartbeat.rs
```diff
@@ -153,7 +153,7 @@ fn maybe_process_response() {
                         }
                     };
 
-                    if let Err(err) = state::insert_block(state, Block::new(block)) {
+                    if let Err(err) = state::insert_block(state, Block::new(block.clone())) {
```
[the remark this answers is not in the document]
dragoljub-duric: Good catch, I do not know why clippy did not complain.

canister/src/heartbeat.rs
```diff
@@ -164,6 +164,17 @@ fn maybe_process_response() {
                         return;
                     }
                 }
+                for block_header_blob in response.next.iter() {
+                    let block_header = BlockHeader::consensus_decode(block_header_blob.as_slice())
+                        .expect("block header decoding must succeed");
```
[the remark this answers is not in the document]
dragoljub-duric: Good catch, I forgot to change this. My idea is when decoding fails, we abort adding to next_block hashes for that response, because if we continue, we may not have the previous block on which to attach.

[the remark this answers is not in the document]
dragoljub-duric: Done. PTAL.

canister/src/unstable_blocks/next_block_hashes.rs
```diff
@@ -0,0 +1,255 @@
...
+        block_hashes.insert(&hash1, 5);
+
+        assert_eq!(block_hashes.get_max_height(), Some(5));
+        assert_eq!(*block_hashes.hash_to_height.get(&hash1).unwrap(), 5);
+        assert_eq!(
+            *block_hashes.height_to_hash.get(&5).unwrap(),
+            vec![hash1.clone()]
+        );
+
+        // Check that insertion the same element does not
```
ielashi, line 101:
```suggestion
        // Check that inserting the same element does not
```

canister/src/unstable_blocks/next_block_hashes.rs
```diff
@@ -0,0 +1,255 @@
...
+        assert_eq!(block_hashes.get_max_height(), Some(5));
+        assert_eq!(*block_hashes.hash_to_height.get(&hash2).unwrap(), 5);
+        assert_eq!(
+            *block_hashes.height_to_hash.get(&5).unwrap(),
+            vec![hash1.clone(), hash2.clone()]
+        );
+    }
+
+    #[test]
+    pub(crate) fn test_remove_block() {
```
ielashi, line 125:
```suggestion
    pub(crate) fn test_remove() {
```

canister/src/lib.rs
```diff
@@ -93,18 +94,21 @@ pub fn get_current_fee_percentiles(
 ) -> Vec<MillisatoshiPerByte> {
     verify_api_access();
     verify_network(request.network.into());
+    verify_fully_synced();
```
ielashi, line 97: I believe we're not testing any of these endpoints that this check is being applied. Maybe we can an add an e2e test that puts the bitcoin canister in a state where it's not synced, and verifies that these API requests get rejected?

canister/src/lib.rs
```diff
@@ -210,6 +214,31 @@ fn verify_api_access() {
     });
 }
 
+/// Used to determine if the canister is synced based on difference
+/// between the maximum height of all block headers and the
+/// maximum height of all unstable blocks
+pub const SYNCING_THRESHOLD: u32 = 2;
```
ielashi, line 220: nit: move constants to the top of the file.

canister/src/unstable_blocks.rs
```diff
@@ -71,6 +77,38 @@ impl UnstableBlocks {
...
+        block_hash: &BlockHash,
+        stable_height: Height,
+    ) {
+        let height = match self.next_block_hashes.get_height(prev_block_hash) {
+            Some(prev_height) => *prev_height,
+            None => {
+                if let Ok(depth) = self.block_depth(prev_block_hash) {
+                    stable_height + depth
+                } else {
+                    return;
```
ielashi, line 101: It looks like we're silently ignoring the case where we are inserting a block hash that doesn't extend the tree. I think it's more clear for this API to return a `BlockDoesNotExtendTree` error and let the caller decide what to do.

canister/src/lib.rs
```diff
@@ -93,18 +94,21 @@ pub fn get_current_fee_percentiles(
...
     verify_api_access();
     verify_network(request.network.into());
+    verify_fully_synced();
     api::get_current_fee_percentiles()
 }
 
 pub fn get_balance(request: GetBalanceRequest) -> Result<Satoshi, GetBalanceError> {
     verify_api_access();
     verify_network(request.network.into());
+    verify_fully_synced();
```
ielashi, line 104: I wonder, should we add a flag in the `Config`, something like `disable_api_if_not_fully_synced: Flag`, that enables/disables this check? I think that'd be useful in case we find issues with this check in production and need a quick way to disable it.

canister/src/unstable_blocks.rs
```diff
@@ -81,7 +119,7 @@ pub fn peek(blocks: &UnstableBlocks) -> Option<&Block> {
 /// Pops the `anchor` block iff ∃ a child `C` of the `anchor` block that
 /// is stable. The child `C` becomes the new `anchor` block, and all its
 /// siblings are discarded.
-pub fn pop(blocks: &mut UnstableBlocks) -> Option<Block> {
+pub fn pop(blocks: &mut UnstableBlocks, stable_height: Height) -> Option<Block> {
```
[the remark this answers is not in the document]
dragoljub-duric: I agree, I have tried without passing `stable_height` but it will be painful.

canister/src/lib.rs
dragoljub-duric: I am not sure how beneficial that will be, because it will require some time to make such tests I am not sure I have any idea how I will put the canister in a not synced state without too much work. And we are already testing the behavior of verify_fully_synced(), hence the only way to make it not work is to delete the check from the methods.  But, if you feel strongly about it, we can add it to the follow-up. This PR is already too long:)

ielashi:
Adding @THLO as well as a reviewer.

canister/src/lib.rs
dragoljub-duric: But testing of `get_current_fee_percentiles`, `get_balance`, and `get_utxos` should be easy. Do you think that will be enough?

canister/src/lib.rs
dragoljub-duric: [Here](https://github.com/dfinity/bitcoin-canister/pull/156) is the follow up with change

canister/src/lib.rs
ielashi: Here's how I think about it:

1. We're adding a new functionality.
2. This functionality, if it's buggy, poses a significant risk.

By the above logic, to me this is already a very strong reason to add test cases. Note that in my logic, I didn't consider how difficult writing the test would be nor how simple the implementation is. These, IMHO, are not the right problems to decide on whether tests should be included or not. If writing tests is painful, that means we need to improve our tooling to make testing easier. And if an implementation is simple enough today, it is not a guarantee that the implementation won't evolve to be more complex in the future.

> I am not sure I have any idea how I will put the canister in a not synced state without too much work.\
It's actually very simple. If you create something like the e2e scenarios where you are returning no blocks and a few `next` block hashes forever, the bitcoin canister will be in a state where it's not synced.

I'm ok with that to be part of a follow-up, but let's create a JIRA ticket for it please so that we don't forget.

dragoljub-duric: [Here](https://dfinity.atlassian.net/browse/EXC-1362) is jira link.

canister/src/lib.rs
```diff
@@ -38,6 +38,11 @@ use std::convert::TryInto;
 use std::{cell::RefCell, cmp::max};
 use utxo_set::UtxoSet;
 
+/// Used to determine if the canister is synced based on difference
+/// between the maximum height of all block headers and the
+/// maximum height of all unstable blocks
+pub const SYNCING_THRESHOLD: u32 = 2;
```
ielashi, line 44: nit: This isn't really a syncing threshold - we're always syncing no matter what. Instead, it's a threshold of whether or not we are synced.

Also, this constant should be private.

How about?

```suggestion
/// The maximum number of blocks the canister can be behind the tip to be considered synced.
const SYNCED_THRESHOLD: u32 = 2;
```

canister/src/heartbeat.rs
```diff
@@ -164,6 +164,41 @@ fn maybe_process_response() {
                         return;
                     }
                 }
+                for block_header_blob in response.next.iter() {
```
ielashi, line 167: nit: consider moving this loop into a method.

ielashi (approved):
LGTM. I updated the description a bit since the document that was linked there isn't available publicly.

canister/src/lib.rs
```diff
@@ -210,6 +217,26 @@ fn verify_api_access() {
     });
 }
 
+/// Verifies that if the difference between the maximum height
+/// of all block headers and the maximum height of all unstable
+/// blocks is at most the SYNCED_THRESHOLD.
+fn verify_fully_synced() {
```
THLO, line 223: Given that we allow a gap, wouldn't it be better to call it `verify_synced`?

dragoljub-duric: Done.

canister/src/unstable_blocks.rs
```diff
@@ -71,6 +77,39 @@ impl UnstableBlocks {
...
+                } else {
+                    return Err(BlockDoesNotExtendTree(block_hash.clone()));
+                }
+            }
+        } + 1;
+        self.next_block_hashes.insert(block_hash, height);
+        Ok(())
+    }
+
+    // Public only for testing purpose.
```
THLO, line 109: So, would `pub(crate)` suffice?

canister/src/unstable_blocks.rs
```diff
@@ -240,7 +288,7 @@ mod test {
         let utxos = UtxoSet::new(network);
         let mut forest = UnstableBlocks::new(&utxos, 1, anchor, network);
```
THLO, line 289: Nit: It is always a connected tree, right? So, calling the variable `forest` is no longer appropriate.

canister/src/unstable_blocks/next_block_hashes.rs
```diff
@@ -0,0 +1,255 @@
...
+
+        assert_eq!(block_hashes.get_max_height(), Some(5));
+        assert_eq!(*block_hashes.hash_to_height.get(&hash1).unwrap(), 5);
+        assert_eq!(
+            *block_hashes.height_to_hash.get(&5).unwrap(),
+            vec![hash1.clone()]
+        );
+
+        // Check that inserting the same element does not
+        // create a duplicates.
```
THLO, line 102:
```suggestion
        // create a duplicate.
```

THLO (approved):

canister/src/unstable_blocks.rs
dragoljub-duric: Done.

PR #154, opened 2023-03-16
feat: use the guard pattern when fetching blocks.

ielashi:
# Problem
If for any reason the logic in fetching blocks panics, we'll forever stop fetching new blocks.

# Solution
Use the guard pattern, which guarantees that we always set the `is_fetching_blocks` flag to `false` and continue fetching new blocks whenever we receive a response, even if it traps.

### DIFF
--- a/canister/src/heartbeat.rs
+++ b/canister/src/heartbeat.rs
@@ -32,12 +32,16 @@ pub async fn heartbeat() {
 // Fetches new blocks if there isn't a request in progress and no complete response to process.
 // Returns true if a call to the `blocks_source` has been made, false otherwise.
 async fn maybe_fetch_blocks() -> bool {
-    if with_state(|s| {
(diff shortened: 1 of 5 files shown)


canister/src/heartbeat.rs
```diff
@@ -49,22 +53,15 @@ async fn maybe_fetch_blocks() -> bool {
...
 
     let response: Result<(GetSuccessorsResponse,), _> =
         call_get_successors(with_state(|s| s.blocks_source), request).await;
 
     print(&format!("Received response: {:?}", response));
 
-    // Release the heartbeat lock and save the response.
+    // Save the response.
     with_state_mut(|s| {
-        s.syncing_state.is_fetching_blocks = false;
```
dragoljub-duric, line 66: Should we here be `guard.drop()`?

PR #156, opened 2023-03-21
chore: Add disable_api_if_not_fully_synced config flag

dragoljub-duric:


canister/src/state.rs
```diff
@@ -50,6 +50,9 @@ pub struct State {
     /// Flag to control access to the apis provided by the canister.
     #[serde(default)]
     pub api_access: Flag,
+
+    /// Flag to control syncing of the next blocks.
```
ielashi, line 54: I don't think this is correct. This flag doesn't effect the syncing of the blocks.

```suggestion
    /// Flag to determine if the API should be automatically disabled if the canister isn't fully synced.
```

canister/src/types.rs
```diff
@@ -44,6 +44,9 @@ pub struct Config {
 
     /// Flag to control access to the apis provided by the canister.
     pub api_access: Flag,
+
+    /// Flag to control syncing of the next blocks.
```
ielashi, line 48: ditto

ielashi (approved):

PR #157, opened 2023-03-22
fix: fix bug in retrieving the caller in the set_config endpoint.

ielashi:
### Problem:
When calling the set_config API, the following error is returned:

```
"ic0_msg_caller_size" cannot be executed in reply callback mode
```

The reason here is that the caller is fetched after making an async request, at which point it becomes unavailable.

### Solution:
Store the caller in a variable before making the async request.


dragoljub-duric:
LGTM.

dragoljub-duric (approved):

PR #158, opened 2023-03-23
chore: Add e2e tests for disable_api_if_not_fully_synced

dragoljub-duric:


e2e-tests/verify-is-synced/src/main.rs
```diff
@@ -0,0 +1,262 @@
...
+            partial_block: BLOCKS.with(|b| b.borrow()[1].clone())[0..20].to_vec(),
+            next: vec![],
+            remaining_follow_ups: 2,
+        })
+    } else if count == 2 {
+        // Send another part of block 2.
+        GetSuccessorsResponse::FollowUp(BLOCKS.with(|b| b.borrow()[1].clone())[20..40].to_vec())
+    } else if count == 3 {
+        // Send rest of block 2.
+        GetSuccessorsResponse::FollowUp(BLOCKS.with(|b| b.borrow()[1].clone())[40..].to_vec())
```
ielashi, line 213: Splitting the block across multiple responses isn't being tested in this scenario, so I'd just return the complete response.

e2e-tests/verify-is-synced/src/main.rs
```diff
@@ -0,0 +1,262 @@
...
+    static BLOCKS: RefCell<Vec<BlockBlob>> = RefCell::new(Vec::new());
+
+    static COUNT: Cell<u64> = Cell::new(0);
+
+    static BLOCK_HEADERS: RefCell<Vec<BlockHeaderBlob>> = RefCell::new(Vec::new());
+}
+
+// Initialize the blocks.
+#[init]
+fn init() {
```
ielashi, line 79: Do we need this relatively complicated logic? Why not just create empty blocks without any transactions?

e2e-tests/verify-is-synced.sh
```diff
@@ -0,0 +1,182 @@
...
+  exit 1
+fi
+
+# Request the current fee percentiles. This is only for profiling purposes.
+dfx canister call bitcoin bitcoin_get_current_fee_percentiles '(record {
+  network = variant { regtest };
+})'
+dfx canister call bitcoin bitcoin_get_current_fee_percentiles '(record {
+  network = variant { regtest };
+})'
```
ielashi, line 180: Do we need to make this call twice?

e2e-tests/verify-is-synced/src/main.rs
dragoljub-duric: If I tried to create block without transactions I get following error:
`Creating UI canister on the local network.
The UI canister on the "local" network is "rrkah-fqaaa-aaaaa-aaaaq-cai"
Installing code for canister e2e-verify-is-synced, with canister ID rwlgt-iiaaa-aaaaa-aaaaa-cai
[Canister rwlgt-iiaaa-aaaaa-aaaaa-cai] Panicked at 'could not initialize thread_rng: All entropy sources failed (permanently unavailable); cause: OsRng: support for wasm32 requires emscripten, stdweb or wasm-bindgen (permanently unavailable)', /home/dduric/.cargo/registry/src/github.com-1ecc6299db9ec823/rand-0.6.5/src/rngs/thread.rs:80:17
Error: Failed while trying to deploy canisters.
Caused by: Failed while trying to deploy canisters.
  Failed while trying to install all canisters.
    Failed to install wasm module to canister 'e2e-verify-is-synced'.
      Failed to install wasm in canister 'rwlgt-iiaaa-aaaaa-aaaaa-cai'.
        Failed to install wasm.
          The Replica returned an error: code 5, message: "Canister rwlgt-iiaaa-aaaaa-aaaaa-cai trapped explicitly: Panicked at 'could not initialize thread_rng: All entropy sources failed (permanently unavailable); cause: OsRng: support for wasm32 requires emscripten, stdweb or wasm-bindgen (permanently unavailable)', /home/dduric/.cargo/registry/src/github.com-1ecc6299db9ec823/rand-0.6.5/src/rngs/thread.rs:80:17"`

dragoljub-duric: But if I put just one transaction per block it works, hence I did that.

e2e-tests/verify-is-synced.sh
```diff
@@ -0,0 +1,139 @@
...
+
+# The address has 2 UTXOs.
+if ! [[ $(num_utxos "$UTXOS") = 2 ]]; then
+  echo "FAIL"
+  exit 1
+fi
+
+dfx canister call bitcoin bitcoin_get_current_fee_percentiles '(record {
+  network = variant { regtest };
+})'
```
ielashi, line 137: I don't think this does anything actually. As in, this will call the endpoint but succeed no matter what. I think your intention here is to check that it's returning some values and not rejecting the request, which you'll need to do explicitly.

dragoljub-duric: It will panic if not synced hence the script will fail.

dragoljub-duric: I have added the check.

ielashi (approved):
LGTM, thanks!

PR #159, opened 2023-03-24
fix: main-state-builder compilation and add it to CI

ielashi:
## Problem
`main-state-builder`, a canister used to bootstrap the Bitcoin state, doesn't compile.

## Solution
* Fix the compilation problem.
* Add the crate to CI to avoid this from happening again.

Closes EXC-1364


dragoljub-duric:
LGTM.

PR #160, opened 2023-03-24
feat: add benchmark for block insertion

ielashi:
## Problem
We currently do some profiling based on some of e2e tests, but these scenarios aren't very realistic. As a consequence, it wasn't noticeable how expensive some operations are, such as block insertion.

## Solution
Introduce some realistic benchmarks. This commit includes a benchmark for block insertion, where we insert the first 300 blocks of the Bitcoin testnet. Currently this consumes over 100B instructions (!!). This commit doesn't include the tooling needed to run this benchmark. That will be included in follow-up commits.

EXC-1366


dragoljub-duric:
Seems LGTM to me, I guess the follow-up will include a script to run the benchmark.

PR #162, opened 2023-03-24
chore: move ic_btc_types and ic_btc_test_utils into repo.

ielashi:
We are currently relying on the `ic_btc_types` and `ic_btc_test_utils` crates from the IC repo. Once Cargo is fully deprecated from the IC repo, we won't be able to import these crates. Additionally, they structurally fit here better than in the IC repo. For example, `ic_btc_test_utils` isn't used in the IC repo at all.

`ic_btc_types` has been renamed in the process to `ic_btc_interface`, since it's actually the public interface of the Bitcoin canister.


dragoljub-duric:
Seems LGTM to me.

PR #167, opened 2023-03-29
feat: add metric to track if the canister is synced

ielashi:
## Problem
There is now logic in the canister to disable its API automatically if it's not synced with the network, but we have no way of tracking whether or not our definition of "synced" is too sensitive, and whether or not we're frequently flapping between the synced and not synced states.

## Solution
Add a `is_synced` metric to the canister that we can use to monitor changes in the synced state.

### DIFF
--- a/canister/src/lib.rs
+++ b/canister/src/lib.rs
@@ -219,7 +219,7 @@ fn verify_api_access() {
     });
 }
 
-// Verifies that if the difference between the maximum height
+// Verifies if the difference between the maximum height
 // of all block headers and the maximum height of all unstable
 // blocks is at most the SYNCED_THRESHOLD.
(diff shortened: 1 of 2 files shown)


dsarlis (approved):

canister/src/lib.rs
```diff
@@ -219,27 +219,34 @@ fn verify_api_access() {
     });
 }
 
-/// Verifies that if the difference between the maximum height
-/// of all block headers and the maximum height of all unstable
-/// blocks is at most the SYNCED_THRESHOLD.
+// Verifies that if the difference between the maximum height
```
dsarlis, line 222:
```suggestion
// Verifies if the difference between the maximum height
```

PR #171, opened 2023-04-03
refactor: clean up interface types.

ielashi:
* Removes types in `canister/src/types.rs` that are already present in `ic_btc_interface`.
* Merges the `Network` and `NetworkSnakeCase` enums, as keeping both of them is redundant.


dragoljub-duric (approved):
LGTM

PR #172, opened 2023-04-03
add ic-http-mock

maksymar:
This PR adds the `ic-http-mock` crate, providing developers with a simplified solution for mocking HTTP outcalls on IC during testing.

Maybe in the future it should be published as a separate crate, but so far placing it here is good enough.


ic-http-mock/src/transform.rs
```diff
@@ -0,0 +1,47 @@
...
+
+/// Creates a `TransformContext` from a transform function and a context.
+/// Also inserts the transform function into a thread-local hashmap.
+pub fn create_transform_context(func: TransformFn, context: Vec<u8>) -> TransformContext {
+    let function_name = get_function_name(func).to_string();
+    registry_insert(function_name.clone(), func);
+
+    TransformContext {
+        function: TransformFunc(candid::Func {
+            principal: Principal::anonymous(),
```
ielashi, line 42: Is the `principal` here the caller of this method? If yes, then that should be the management canister, according to the interface spec.

ic-http-mock/Cargo.toml
```diff
@@ -0,0 +1,29 @@
+[package]
+name = "ic-http-mock"
+description = "Mocking HTTPS Outcalls on the Internet Computer"
+version = "0.1.0"
+edition = "2021"
+authors = ["Maksym Arutyunyan"]
+repository = "https://github.com/TODO/ic-http-mock" # TODO: Update this
+homepage = "https://github.com/TODO/ic-http-mock"   # TODO: Update this
```
ielashi, line 8: Will these be updated before this PR is merged? If not, I'd remove them for now.

ic-http-mock/example_canister/src/canister_backend/Cargo.toml
```diff
@@ -0,0 +1,23 @@
...
+ic-cdk-macros = "0.6.0"
+ic-http-mock = {path = "../../../"}
+serde = { version = "1.0.158", features = [ "derive" ] }
+serde_json = "1.0.94"
+
+[target.'cfg(not(target_arch = "wasm32"))'.dependencies]
+tokio = { version = "1.15.0", features = [ "full" ] }
+
+[dev-dependencies]
+tokio = { version = "1.15.0", features = [ "full" ] }
```
ielashi, line 23: Are you sure this is needed? The dependency is already declared as a non-dev dependency, and we can't include tokio I believe in the wasm build, which would make this redundant.

ic-http-mock/example_canister/src/canister_backend/Cargo.toml
```diff
@@ -0,0 +1,23 @@
...
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[lib]
+crate-type = ["cdylib"]
+
+[dependencies]
+candid = "0.8.2"
+ic-cdk = "0.6.0"
+ic-cdk-macros = "0.6.0"
+ic-http-mock = {path = "../../../"}
```
ielashi, line 15: nit: It's odd to me that a mocking library is included as dependency of the canister. I would expect mocking libraries only to be included as dev-dependencies.

ic-http-mock/src/mock.rs
```diff
@@ -0,0 +1,134 @@
...
+
+/// Returns the number of times the given request has been called.
+/// Returns 0 if no mock has been found for the request.
+pub fn times_called(request: &CanisterHttpRequestArgument) -> u64 {
+    get(request).map(|mock| mock.times_called).unwrap_or(0)
+}
+
+/// Create a hash from a CanisterHttpRequestArgument, which includes its URL,
+/// method, headers, body, and optionally, its transform function name.
+fn hash(request: &CanisterHttpRequestArgument) -> String {
```
ielashi, line 112: This exists because `Hash` isn't derived for `CanisterHttpRequestArgument`, right?

ielashi (commented):
Great work, as always, @maksymar :) I appreciate your attention to detail here.

The code looks great. I have some high-level comments:

1. The name of this crate, `ic-http-mock` suggests that it's a mocking library. In practice though, this isn't exactly a mocking library since canisters need to include it as a dependency (as opposed to a dev-dependency) to use it. Would it be more correct to say that crate offers a simplified API to make http requests, with the additional feature that these can be mocked?

Ideally, if we want this to be a separate crate, the mocking part of the code would be protected by a feature flag. That way, a canister that just wants to use this library's nice HTTP API can use it, without having to depend on mocking code in production.

2. Some of the improvements in this library also make sense for them to be put in the `ic-cdk`, such as the request/response builders. I'd suggest moving as much of that as possible into `ic-cdk`. It might even make sense to move the entirety of this logic into `ic-cdk` in the future.

Given the timing, would it make sense to keep this code as something internal to the `watchdog` canister, and then we can afterwards iterate towards making it a separate crate with proper feature flags and/or moving some of the logic into `ic-cdk`?

maksymar:
> Great work, as always, @maksymar :) I appreciate your attention to detail here.

Thank you! I'm glad you like it :)

> The code looks great. I have some high-level comments:
> 
> 1. The name of this crate, `ic-http-mock` suggests that it's a mocking library. In practice though, this isn't exactly a mocking library since canisters need to include it as a dependency (as opposed to a dev-dependency) to use it. Would it be more correct to say that crate offers a simplified API to make http requests, with the additional feature that these can be mocked?
> 
> Ideally, if we want this to be a separate crate, the mocking part of the code would be protected by a feature flag. That way, a canister that just wants to use this library's nice HTTP API can use it, without having to depend on mocking code in production.

Yes, that makes perfect sense. So, I'll rename `ic-http-mock` to just `ic-http` with two groups of methods:

1. API for creating requests/responses and sending those requests under `ic_http::*`
2. API for mocking under `ic_http::mock::*`, that'll include `mock` and `times_called`.

> 2. Some of the improvements in this library also make sense for them to be put in the `ic-cdk`, such as the request/response builders. I'd suggest moving as much of that as possible into `ic-cdk`. It might even make sense to move the entirety of this logic into `ic-cdk` in the future.
> 
> Given the timing, would it make sense to keep this code as something internal to the `watchdog` canister, and then we can afterwards iterate towards making it a separate crate with proper feature flags and/or moving some of the logic into `ic-cdk`?

I agree, let's do this gradually, first adding it here, then potentially moving as much as possible into `ic-cdk`.

Thank you @ielashi for such a thoughtful review!

ic-http-mock/src/mock.rs
```diff
@@ -0,0 +1,134 @@
...
+pub fn mock(request: &CanisterHttpRequestArgument, response: &HttpResponse) {
+    mock_with_delay(request, response, Duration::from_secs(0));
+}
+
+/// Adds a mock for a given HTTP request and response. The mock will be returned by
+/// subsequent calls to `http_request` that match the same request. The response will be
+/// returned after a delay specified by the `delay` argument.
+pub fn mock_with_delay(
+    request: &CanisterHttpRequestArgument,
+    response: &HttpResponse,
```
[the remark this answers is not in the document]
maksymar: Done.

ic-http-mock/src/mock.rs
```diff
@@ -0,0 +1,134 @@
...
+        // TODO: Use a non-blocking sleep when the CDK supports it.
+        #[cfg(not(target_arch = "wasm32"))]
+        tokio::time::sleep(mock.delay).await;
+    }
+
+    // Check if the response body exceeds the maximum allowed size.
+    if let Some(max_response_bytes) = mock.request.max_response_bytes {
+        if mock.response.body.len() as u64 > max_response_bytes {
+            return Err((
+                RejectionCode::SysFatal,
```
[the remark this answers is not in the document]
maksymar: I tried to copy the same error that IC has, which is `SysFatal` error in this case. See [code](https://sourcegraph.com/github.com/dfinity/ic@271b367d7c8d7d3d3d35505bdf0dbe77ff01ab65/-/blob/rs/https_outcalls/adapter/src/rpc_server.rs?L247).

ic-http-mock/src/request.rs
```diff
@@ -0,0 +1,90 @@
...
+    CanisterHttpRequestArgument, HttpHeader, HttpMethod, TransformContext,
+};
+
+/// Creates a new `HttpRequestBuilder` to construct an HTTP request.
+pub fn create_request() -> HttpRequestBuilder {
+    HttpRequestBuilder::new()
+}
+
+/// Represents a builder for an HTTP request.
+pub struct HttpRequestBuilder {
```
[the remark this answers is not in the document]
maksymar: Noted.

ic-http-mock/src/transform.rs
maksymar: Done.

ic-http-mock/Cargo.toml
maksymar: Done.

ic-http-mock/example_canister/src/canister_backend/Cargo.toml
maksymar: Done.

ic-http-mock/example_canister/src/canister_backend/Cargo.toml
maksymar: Done, renamed `ic-http-mock` to `ic-http` which provides a simple http outcall API with and extra convenience of mocking it in tests.

ic-http-mock/src/mock.rs
maksymar: Yes, that's exactly right.

maksymar:
Applied the changes from the above, except maybe hiding mocking http_request behind a feature flag, but rather leaving it platform related. 

PTAL.

ielashi:
LGTM, thanks! As a follow-up, I would suggest moving the crate inside the `watchdog` folder to make it clear that, for now, it's specific to the watchdog canister until the changes are moved to `ic-cdk`.

ielashi (approved):

ic-http/src/lib.rs
```diff
@@ -0,0 +1,165 @@
...
+//! ```
+//!
+
+mod http_request;
+mod request;
+mod response;
+mod transform;
+
+// Export.
+pub mod mock;
```
ielashi, line 121: nit: Consider not importing this module when compiling for wasm32.

maksymar: Done.

maksymar:
> LGTM, thanks! As a follow-up, I would suggest moving the crate inside the `watchdog` folder to make it clear that, for now, it's specific to the watchdog canister until the changes are moved to `ic-cdk`.

Noted, thank you!

PR #175, opened 2023-04-04
fix: next block headers validation

dragoljub-duric:
Problem: 
The code for inserting the next [block header validation](https://github.com/dfinity/bitcoin-canister/blob/master/canister/src/heartbeat.rs#L145) is only checking proof-of-work. That’s only one of several checks. 

Solution: 
Instead, we should be validating the block header with [this method](https://github.com/dfinity/bitcoin-canister/blob/master/validation/src/header.rs#L58).


canister/src/validation.rs
```diff
@@ -1,11 +1,11 @@
-use crate::{blocktree::BlockDoesNotExtendTree, state::State, unstable_blocks, Block};
+use crate::{blocktree::BlockDoesNotExtendTree, state::State, unstable_blocks};
 use bitcoin::BlockHeader;
 use ic_btc_validation::HeaderStore;
 
 /// A structure passed to the validation crate to validate a specific block header.
 pub struct ValidationContext<'a> {
     state: &'a State,
-    chain: Vec<&'a Block>,
+    chain: Vec<(&'a BlockHeader, crate::types::BlockHash)>,
```
dragoljub-duric, line 8: I needed this to use already existing code in ValidationContext, and evermore BlockHeader is sufficient to be stored.  BlockHash is stored to avoid unnecessary recomputation of hash each time we are using BlockHeader.

canister/src/validation.rs
```diff
@@ -16,20 +16,47 @@ impl<'a> ValidationContext<'a> {
...
 
         Ok(Self { state, chain })
     }
+
+    fn insert_in_chain(&mut self, item: (&'a BlockHeader, crate::types::BlockHash)) {
+        self.chain.push(item)
+    }
+    /// Initialize a `ValidationContext` for the given block header.
+    /// The given block header can be in the 'NextBlockHeaders'.
+    pub fn new_with_next_block_headers(
```
dragoljub-duric, line 32: This method just creates ValidationContex with a chain field extended for a blocks in next_block_headers.

canister/src/unstable_blocks/next_block_headers.rs
```diff
@@ -0,0 +1,291 @@
+use crate::types::BlockHash;
+use bitcoin::BlockHeader;
+use ic_btc_interface::Height;
+use serde::{Deserialize, Serialize};
+use std::collections::BTreeMap;
+
```
dragoljub-duric, line 6: This file is renamed, with a small refactoring (now instead of `hash_to_height: BTreeMap<BlockHash, Height>` we have     `hash_to_height_and_header: BTreeMap<BlockHash, (Height, BlockHeader)>` in `NextBlockHeaders and some fields and comments are renamed).

canister/src/validation.rs
ielashi: Is there a reason to not be extending the chain with the next block headers? i.e. why not use this contructor everytime and delete the the `new` method?

canister/src/validation.rs
```diff
@@ -16,20 +16,47 @@ impl<'a> ValidationContext<'a> {
...
+            .iter()
+            .map(|block| (block.header(), block.block_hash()))
+            .collect();
 
         Ok(Self { state, chain })
     }
+
+    fn insert_in_chain(&mut self, item: (&'a BlockHeader, crate::types::BlockHash)) {
+        self.chain.push(item)
+    }
```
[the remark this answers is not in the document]
dragoljub-duric: Done.

canister/src/validation.rs
dragoljub-duric: Counterexample for not using the `new_with_next_block_headers` everywhere:
If we have two blocks `A` and `B`, and the previous block of `B` is `A`, and we receive only `A`'s header in the next block headers followed by receiving complete block `B`. If we use this method, validation will pass because it will find `B`'s predecessor(`A`) in `chain`, but it should fail because we still did not receive the whole block `A`.

canister/src/validation.rs
dragoljub-duric: I was thinking about that, that is a good idea.

dragoljub-duric: I will leave this as it is, and change it in the follow-up. Do you want me to create a new type or you will do it?

ielashi, line 8: nit: Please add a comment as to why we're storing both the has and header, and maybe add a jira ticket as well for us to clean this up later.

ielashi (approved):

PR #176, opened 2023-04-06
chore: upgrade stable structures to version 0.5.2

ielashi:
The API for stable structures changes quite significantly between the two versions, so the change is relatively involved.

Due to a (now known) performance regression in stable structures >= 0.4.0, `address_utxos` have been migrated to use a `Blob` in the `BTreeMap` rather than an `AddressUtxo`. Once the performance regression is fixed, we'll use `AddressUtxo` as the type again.

There's a minor performance regression, mostly due to the conversions between `Blob` and `Vec`. These will be addressed once the regression mentioned above is addressed, but given that it's minor, it's not a blocker for the upgrade.


canister/src/types.rs
```diff
@@ -355,21 +353,85 @@ impl StableStructuresStorable for AddressUtxo {
...
+        let (start_height, start_outpoint) = match utxo {
+            Some(utxo) => (utxo.height, utxo.outpoint.clone()),
+
+            // No UTXO specified. Start with the minimum value possible for a height and OutPoint.
+            // Heights are sorted in descending order, so u32::MAX is considered its minimum.
+            None => (u32::MAX, OutPoint::new(Txid::from(vec![0; 32]), 0)),
+        };
+
+        // The end of the range is the maximum value possible for a height and OutPoint.
+        // i.e. The range with matches with all UTXOs of that address that are >= the given UTXO.
```
dsarlis, line 395:
```suggestion
        // i.e. the range that matches with all UTXOs of that address that are >= the given UTXO.
```

dsarlis (approved):
> There's a minor performance regression, mostly due to the conversions between Blob and Vec

Something that I thought about while reviewing and related to the small performance regression: it would be nice to see the diff in benchmarks in percentages to more easily confirm the regression is negligible.

ielashi:
> > There's a minor performance regression, mostly due to the conversions between Blob and Vec
> 
> Something that I thought about while reviewing and related to the small performance regression: it would be nice to see the diff in benchmarks in percentages to more easily confirm the regression is negligible (instruction numbers are quite large in nature which makes it harder to eyeball things).

Agreed. The `stable-structures` crate is now using `criterion` to display the output, and that's much easier to eyeball. I think we can do the same here.

PR #177, opened 2023-04-11
feat: create a watchdog canister to fetch Bitcoin block data from explorers

maksymar:
This PR introduces a watchdog canister for tracking bitcoin canister health, which is capable of fetching bitcoin block data from different remote APIs.

The implementation uses the `BitcoinBlockApi` enum variant to describe each remote API provider, along with the corresponding `fetch_data()` function that may require sending one or several HTTP requests. The `Endpoint` enum variant is used to describe each HTTP request, which allows for storing all configuration in one place, including the URL, the transform function exposed by the canister, and its respective implementation. This approach allows for all of the canister's exposed methods to be located at the top level in lib.rs, while the actual implementation of the transform function is stored next to the URL and tests to improve readability.

The implementation uses `ic-http` crate, which allowed for testing API responses on mocked data.

Since this PR is already quite large, it is separated from processing and serving the results to the bitcoin_canister, which will follow in separate PRs.


watchdog/src/endpoints.rs
```diff
@@ -0,0 +1,353 @@
...
+use regex::Regex;
+use serde_json::json;
+
+/// Bitcoin Block API endpoints.
+/// Allows to store the configuration of each endpoint, which includes:
+/// - URL
+/// - Transform function canister endpoint
+/// - Transform function inner implementation to be called inside the canister's endpoint
+#[derive(Debug)]
+pub enum Endpoint {
```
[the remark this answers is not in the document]
maksymar: 1. I wanted to move from separate modules/files to just one for all the endpoints. 
    - Because there are not so many (5-6)
    - Those are unique instances of `HttpRequestConfig` that take just a few input values (url, exposed method, inner transform implementation function)
    - I suppose it improves readability to have all the constructors and tests next to each other (since they look very much the same with slight differences)
  
2. Enum was a good enough choice, since it provides a mapping of a unique name to a specific `HttpRequestConfig`. Its URL is also unique, but it's a bit too long and it's a string -- prone to typo-errors.
    - it can also be a number of free functions, then the function name would be a unique config name.
    
what do you think about switching to just functions? Eg.
```rust
endpoint_api_bitaps_com_block()
endpoint_api_blockchair_com_block()
endpoint_api_blockcypher_com_block()
endpoint_bitcoin_canister()
endpoint_blockchain_info_hash()
endpoint_blockchain_info_height()
endpoint_blockstream_info_hash()
endpoint_blockstream_info_height()
endpoint_chain_api_btc_com_block()
```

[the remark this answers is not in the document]
maksymar: switched to just functions, PTAL.

maksymar:
UPD: 

I've found out that there was a bug in `ic-http` implementation that did not allow to properly test mocked responses with transform functions if they were called concurrently. 

Fixed it in #180, would be nice to merge a fix first and then come back to this PR. 

PTAL.

maksymar:
Thank you @ielashi for reviewing #180! This PR is updated and the 'concurrent http requests` issue is fixed.

PTAL.

watchdog/src/http.rs
```diff
@@ -0,0 +1,67 @@
...
+            request: create_request(url, transform_endpoint),
+            transform_implementation,
+        }
+    }
+
+    // URL is added only for tests.
+    #[allow(dead_code)]
+    pub fn url(&self) -> &str {
+        &self.url
+    }
```
ielashi, line 40: nits:
1. If something is needed for tests, it's clearer to use the `cfg(test)` directive as opposed to marking as dead code.
2. Do you actually need to store the url? Does this suffice?

```
    #[cfg(test)]
    pub fn url(&self) -> &str {
        &self.request.url
    }
```

watchdog/src/main.rs
```diff
@@ -0,0 +1,5 @@
+// Makes exported items available to the crate root.
```
ielashi, line 1: nit: I'd recommend using the same pattern as we do in the bitcoin canister. There, we define the endpoints in `main.rs`, not in `lib.rs`.

ielashi (approved):

watchdog/README.md
```diff
@@ -0,0 +1,18 @@
+# bitcoin-canister-watchdog
+
+TODO: add description.
```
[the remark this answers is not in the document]
maksymar: Done.

watchdog/src/http.rs
maksymar: Great idea, thank you! Done.

watchdog/src/main.rs
maksymar: I started with this approach, but the problem is that functions declared in `main.rs` are not visible in the underlying crates, which is needed to register transformation functions. That's why we have a compromise here: all endpoints are declared in `lib.rs`, which makes transformation functions visible downstream and makes them available to register the requests.

PR #180, opened 2023-04-12
fix: fix mocking concurrent http requests with transform functions in ic-http

maksymar:
This PR fixes storing transform functions in a thread-local storage which was preventing to test concurrent http requests.

Previously it was incorrectly calculating the name of the function, since it was called inside `create_transform_context` function with a specific type of `TransformFn`, which loses the name of the function (in contrast to a template parameter) and only stores the function signature (which is always the same).

To fix this it had to convert `create_transform_context` to a template and store a boxed function pointer in a thread-local storage. Added tests to make sure function names are stored properly now.

### DIFF
--- a/ic-http/src/lib.rs
+++ b/ic-http/src/lib.rs
@@ -125,23 +125,29 @@ pub mod mock;
 pub use crate::http_request::http_request;
 pub use crate::request::create_request;
 pub use crate::response::create_response;
-pub use crate::transform::TransformFn;
 
-use crate::mock::{hash, Mock};
-use ic_cdk::api::management_canister::http_request::CanisterHttpRequestArgument;
-use std::collections::HashMap;
-use std::sync::RwLock;
+#[cfg(not(target_arch = "wasm32"))]
+use {
+    crate::mock::{hash, Mock},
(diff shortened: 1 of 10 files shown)


ic-http/src/lib.rs
```diff
@@ -125,42 +125,67 @@ pub mod mock;
 pub use crate::http_request::http_request;
 pub use crate::request::create_request;
 pub use crate::response::create_response;
-pub use crate::transform::TransformFn;
 
-use crate::mock::{hash, Mock};
-use ic_cdk::api::management_canister::http_request::CanisterHttpRequestArgument;
-use std::collections::HashMap;
-use std::sync::RwLock;
+#[cfg(not(target_arch = "wasm32"))]
```
ielashi, line 129: Almost everything in this file is under the `#[cfg(not(target_arch = "wasm32"))]` directive. Consider moving that entire logic into a separate module so that we don't need to specify this directive for every line.

ielashi (approved):

ic-http/src/lib.rs
maksymar: Done, thank you, Islam!

PR #182, opened 2023-04-13
feat: add serving health status to watchdog canister 

maksymar:
This PR adds serving bitcoin canister health status calculated on fetched data.

Health status contains: current bitcoin canister height, number of explorers inspected (and their heights), calculated target height (currently a median), height difference and a status code.
Status code can be: no data, ok (height is within the limits), behind and ahead.

Additionally cleaned the code from panics during fetching data, which was prohibitedly strict and failed on any non-200 response status. Added tests to cover that.

Moved all the constant settings into a separate file config.rs, in the future might be configurable on start or during running.

Currently health status is calculated based on the recent data fetch. One possible improvement against accidental explorer hiccups is to add a time-to-live in storage and keep the data for some time. Not critical, since we have 5 explorers at the moment with requirement of at least 3 to get the target height.


watchdog/candid.did
```diff
@@ -1,3 +1,19 @@
+type status_code = variant {
```
dsarlis, line 1: nit: It might be good idea to leave some comments about what the fields represent. Since this is the interface of the canister, it's nice if you can quickly get a high level idea without reading the implementation (e.g. the `other_number` is very hard to know what it is without reading further).

watchdog/src/bitcoin_block_apis.rs
```diff
@@ -1,25 +1,45 @@
...
+#[derive(Debug, Clone, Eq, PartialEq, Hash)]
 pub enum BitcoinBlockApi {
-    // TODO: investigate why this API is not working.
-    #[allow(dead_code)]
     ApiBitapsCom,
-
     ApiBlockchairCom,
     ApiBlockcypherCom,
-    BitcoinCanister,
+    BitcoinCanister, // Not an explorer.
```
dsarlis, line 11: Yeah this part is a bit weird. I would actually expect that we can make a simple inter-canister call to get the height from the bitcoin canister (of course we need to add a method that returns it first). I don't think we should take action immediately but the fact we have to add a comment to clarify is a bit of a code smell.

watchdog/src/fetch.rs
```diff
@@ -0,0 +1,184 @@
+use crate::bitcoin_block_apis::BitcoinBlockApi;
+
+/// The data fetched from the external bitcoin block APIs.
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct BlockInfo {
+    pub provider: BitcoinBlockApi,
```
dsarlis, line 6: nit: Some doc comments on the fields would be nice.

watchdog/src/health.rs
```diff
@@ -0,0 +1,204 @@
...
+/// Compares the source with the other explorers.
+fn compare(source: Option<BlockInfo>, other: Vec<BlockInfo>) -> HealthStatus {
+    let source_height = source.and_then(|block| block.height);
+    let heights = other
+        .iter()
+        .filter_map(|block| block.height)
+        .collect::<Vec<_>>();
+    let other_number = heights.len() as u64;
+    let other_heights = heights.clone();
+    let target_height = if heights.len() < crate::config::MIN_EXPLORERS {
```
dsarlis, line 50:
```suggestion
    let target_height = if other_number < crate::config::MIN_EXPLORERS {
```

watchdog/src/health.rs
```diff
@@ -0,0 +1,204 @@
...
+}
+
+/// The median of the given values.
+fn median<T: std::cmp::Ord + Clone>(mut values: Vec<T>) -> Option<T> {
+    let n = values.len();
+    if n == 0 {
+        None
+    } else {
+        values.sort();
+        let mid = if n % 2 == 0 { (n - 1) / 2 } else { n / 2 };
```
dsarlis, line 85: Technically, the definition of median for an even list of numbers is the average of the two middle values. Is there a reason not using this definition? I don't believe it matters much but just noticed it.

watchdog/src/health.rs
```diff
@@ -0,0 +1,204 @@
...
+    pub other_number: u64,
+    pub other_heights: Vec<u64>,
+    pub target_height: Option<u64>,
+    pub height_diff: Option<i64>,
+    pub status: StatusCode,
+}
+
+/// Calculates the health status of the Bitcoin canister.
+pub fn calculate() -> HealthStatus {
+    compare(
```
dsarlis, line 32: Not sure having the `compare` as an extra function provides a lot of benefit. I'd keep only `calculate` (and maybe rename to `get_health_status`).

watchdog/src/health.rs
```diff
@@ -0,0 +1,204 @@
+use crate::bitcoin_block_apis::BitcoinBlockApi;
+use crate::fetch::BlockInfo;
+use candid::CandidType;
+use serde::{Deserialize, Serialize};
+
+/// Status codes for the health of Bitcoin canister.
+#[derive(Clone, Debug, CandidType, PartialEq, Eq, Serialize, Deserialize)]
+pub enum StatusCode {
+    #[serde(rename = "no_data")]
+    NoData,
```
dsarlis, line 10: I'm thinking that `undefined` or something along those lines would be more appropriate. Or `NotEnoughData`. Because the current one makes it seem like we couldn't get anything back but in the `HealthStatus` we might still see some (but not enough) data.

dsarlis (commented):
A general question: have we setup the [panic hook](https://github.com/dfinity/cdk-rs/blob/main/src/ic-cdk/src/printer.rs#L5) for this canister? It's useful in case there's still some panic because we would get some context printed automatically.

watchdog/src/health.rs
maksymar: A separate `compare` is needed to easily unit test different cases without getting into storage read/write.

watchdog/src/health.rs
maksymar: Done.

watchdog/src/health.rs
maksymar: Done.

watchdog/src/fetch.rs
maksymar: Done.

watchdog/candid.did
maksymar: Done.

maksymar:
> maybe the idea is the bitcoin canister will poll this endpoint periodically and if it's not ok, then it can take actions?

Yes, that was my intention. But we can discuss that further if needed.

dsarlis (approved):
> Yes, that was my intention. But we can discuss that further if needed.

No, it's ok given we have a separate watchdog canister.

PR #183, opened 2023-04-13
feat: add mocking errors to ic-http

maksymar:
This PR adds the possibility of mocking canister rejection errors to ic-http.

Currently it's only possible to mock responses when calling `ic_http::http_request(...).await`, but with this change it'd be possible to mock canister rejection errors.

Additionally simplify example_canister code, add more tests for errors.


ic-http/src/mock.rs
```diff
@@ -8,7 +8,8 @@ use std::time::Duration;
 #[derive(Clone)]
 pub(crate) struct Mock {
     pub(crate) request: CanisterHttpRequestArgument,
-    response: HttpResponse,
+    response: Option<HttpResponse>,
+    error: Option<(RejectionCode, String)>,
```
dsarlis, line 12: How about we have a single field `result` which would be `Option<Result<HttpResponse, (RejectionCode, String)>>`. The reason is that this way it's more clear that it can be either one or another and it's enforced through the type system.

ic-http/src/transform.rs
```diff
@@ -83,4 +84,48 @@ mod test {
         // Assert.
         assert_eq!(names, vec!["transform_function_1", "transform_function_2"]);
     }
+
+    /// Transform function which is intentionally creates a new request
+    /// with itself as a transform function.
```
dsarlis, line 89: This sentence is a bit hard to parse. Maybe rewrite it a bit? Something like:
```
Transform function which intentionally creates a new request passing itself as the target transform function.
```

ic-http/src/transform.rs
```diff
@@ -83,4 +84,48 @@ mod test {
...
+        arg.response
+    }
+
+    /// Creates a request with a transform function which overwrites itself.
+    fn create_request_with_transform() -> CanisterHttpRequestArgument {
+        crate::request::create_request()
+            .url("https://www.example.com")
+            .transform_func(transform_function_with_overwrite, vec![])
+            .build()
+    }
```
dsarlis, line 104: Actually, reading this function, how about we move `args.response` from above inside here directly? Not sure we gain much by having the extra `transform_function_with_overwrite`.

dsarlis (approved):
Looks good. Left some minor comments which I don't think are blocking.

dsarlis:
> Looks good. Left some minor comments which I don't think are blocking.

Ah I didn't realise you had set it to be merged automatically on approval. Feel free to address in a follow-up if you agree with the suggestions.

ic-http/src/mock.rs
maksymar: Done in #185, good idea, thank you!

ic-http/src/transform.rs
maksymar: This has to be a separate function with a name so it can be stored in a thread-local storage by it's name.

ic-http/src/transform.rs
maksymar: Done in #185.

maksymar:
> > Looks good. Left some minor comments which I don't think are blocking.
> 
> Ah I didn't realise you had set it to be merged automatically on approval. Feel free to address in a follow-up if you agree with the suggestions.

I was playing around with this button and forgot to turn it off, sorry about that!
Thank you for the review, I addressed your comments in #185, PTAL.

PR #185, opened 2023-04-14
chore: refactoring storing mocked responses for ic-http

maksymar:
This PR improves storing mocked response to be either result or an error.

This is a follow-up PR to address code review comments from #183.


ic-http/src/mock.rs
```diff
@@ -91,10 +88,11 @@ pub(crate) async fn http_request(
     }
 
     // Return the error if one is specified.
```
dsarlis, line 90: I guess this comment might need a small adjustment? Or just move it in the case where we have an error in the match?

dsarlis (approved):

PR #187, opened 2023-04-18
feat: add watchdog canister config

maksymar:
This PR adds the option to read a watchdog canister's configuration.

Adds a `get_config` endpoint and corresponding bash-tests.


watchdog/tests/get_config.sh
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env bash
```
ielashi, line 1: I think you want to add these tests to the github workflow file. Otherwise, they won't run in CI.

maksymar: Done, thank you!

ielashi (approved):

PR #188, opened 2023-04-18
chore: add panic hook to watchdog canister

maksymar:
This PR adds a panic hook to improve debugging panic with some extra context.

This is a follow-up on the code review from [this comment](https://github.com/dfinity/bitcoin-canister/pull/182#pullrequestreview-1385272013).


watchdog/src/lib.rs
```diff
@@ -23,9 +26,17 @@ thread_local! {
     static BLOCK_INFO_DATA: RwLock<HashMap<BitcoinBlockApi, BlockInfo>> = RwLock::new(HashMap::new());
 }
 
+static START: Once = Once::new();
+
 /// This function is called when the canister is created.
 #[ic_cdk_macros::init]
 fn init() {
+    // Setup the stdlib hooks.
+    START.call_once(|| {
```
dsarlis, line 35: I don't think we need to use the `Once` primitive here. The canister's `init` method will only be invoked once at canister initialization anyway. We should also setup the hook in the `post_upgrade` method (so if the canister is upgraded we don't lose it).

watchdog/src/printer.rs
```diff
@@ -0,0 +1,36 @@
...
+pub fn print(msg: &str) {
+    #[cfg(target_arch = "wasm32")]
+    ic_cdk::api::print(msg);
+
+    #[cfg(not(target_arch = "wasm32"))]
+    println!("{}", msg);
+}
+
+/// Sets a custom panic hook, uses debug.trace
+pub fn set_panic_hook() {
```
dsarlis, line 13: Personal opinion: having to copy this because you want to use your own print function seems like an overkill. Maybe just use the cdk version but guard the code with `#[cfg(target_arch = "wasm32")]` where you need to install the hook? I guess it mainly makes sense for running in a canister anyway, not in a native unit test.

watchdog/src/lib.rs
maksymar: Done. `init` method is called from `post_upgrade` method so it's handled automatically.

ielashi:
It's not clear to me what the panic hook here actually adds. I just tried panicking in this canister with and without this change.

Without the change:

```
          The Replica returned an error: code 5, message: "Canister rwlgt-iiaaa-aaaaa-aaaaa-cai trapped explicitly: Panicked at 'this is a panic', watchdog/src/lib.rs:33:5"
```

With the change:

```
          The Replica returned an error: code 5, message: "Canister rwlgt-iiaaa-aaaaa-aaaaa-cai trapped explicitly: Panicked at 'this is a panic', watchdog/src/lib.rs:37:5"
```

maksymar:
> It's not clear to me what the panic hook here actually adds.

Great observation, Islam. TBH I didn't check initially, trusted the idea that there must have been a reason to use it if it was added into [ic-cdk](https://github.com/dfinity/cdk-rs/blob/main/src/ic-cdk/src/printer.rs#L5).

Just double checked myself, you're right, the panic messages are identical with and without this panic hook.

Thinking out loud about the reasons to use a custom panic hook:

- A custom error reporting, maybe adding extra info, eg. application state, and report panic to monitoring or logging. Not the case.
- Custom error message. Make it more informative and user-friendly. Not the case, I suppose it's informative enough as is.
- Custom panic handling strategy, eg. gracefully shutting down, restarting the canister or other recovery actions. Not the case either.

I'd be glad to see a compelling reason to keep it. Please provide if you have one.

Otherwise I suggest to drop this PR and have a watchdog canister without a custom panic hook. WDYT?

maksymar:
> I'd be glad to see a compelling reason to keep it. Please provide if you have one.
> 
> Otherwise I suggest to drop this PR and have a watchdog canister without a custom panic hook. WDYT?

@dsarlis could you please take a look?

dsarlis:
The main benefit I think is the error message returned to clients of the canister. Without it, you'd see something like "trapped: unreachable" which is not really helpful, whereas with the hook the canister calls ic0.trap explicitly and has the chance to attach a more custom error message.

It is true that when you look at the replica logs they would look very similar (of course you can add more custom ways to report such errors as Maksym pointed out).

It was in any case a suggestion/nice to have (hence not blocking the other PR) so I am fine either way.

maksymar:
Thanks for the clarification, Dimitris. 
I conclude that overall it's nice to have such a hook and since it's already been implemented I'd go with merging it.

Any LGTMs please? :)

watchdog/src/lib.rs
```diff
@@ -35,6 +38,11 @@ thread_local! {
 /// This function is called when the canister is created.
 #[init]
 fn init() {
+    // Setup the stdlib hooks.
+    #[cfg(target_arch = "wasm32")]
+    printer::hook();
```
dsarlis, line 43: If we do this, doesn't it mean we can just use the version from cdk? Why do we still need to copy the code into this printer module?

ielashi:
@maksymar Is this PR still relevant, or should it be closed?

watchdog/src/lib.rs
maksymar: The hook from cdk is not a public method, it's used inside `setup` method behind the flag. 
I don't fully understand the effect of such a flag, especially because in bitcoin canister the hook has to be set up both in init and in post_upgrade methods. To make sure it's set up both times I suppose it's better to have our own copy without a flag.

ielashi:
@maksymar I have yet to find an experiment that shows that this change does anything. If so, should we close this PR?

maksymar:
SGTM!

PR #194, opened 2023-04-25
chore: add tests for explorer api abuse to watchdog canister

maksymar:
This PR adds a test case when API provider responds with a text string instead of a valid JSON.

### DIFF
--- a/watchdog/src/bitcoin_block_apis.rs
+++ b/watchdog/src/bitcoin_block_apis.rs
@@ -1,4 +1,5 @@
 use crate::endpoints::*;
+use crate::print;
 use candid::CandidType;
 use ic_cdk::api::management_canister::http_request::HttpResponse;
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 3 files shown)


ielashi:
I don't understand how the title and description of this PR relate to the actual change. The changes that I see are making adjustments to logs and adding tests. How does that fix how we handle API abuse?

maksymar:
> I don't understand how the title and description of this PR relate to the actual change. The changes that I see are making adjustments to logs and adding tests. How does that fix how we handle API abuse?

good catch, thank you! I must have been tired yesterday when debugging watchdog canister. switched error print messages to panics for debugging purposes and saw a reject message about abusing explorer api (also used 5s ping rate instead of a default 60s for debugging) and incorrectly concluded that getting a string instead of a proper JSON was causing the failure.

still I think it brings value to add such a test case for checking a correct handling of an unexpected text instead of a proper JSON. updated the PR title.

PR #195, opened 2023-04-26
fix: watchdog canister running out of cycles when sending http_request

maksymar:
Attach zero cycles to `http_request` in order not to run out of cycles when running on a system subnet.


ic-http/src/http_request.rs
```diff
@@ -15,5 +15,5 @@ pub async fn http_request(arg: CanisterHttpRequestArgument) -> CallResult<(HttpR
 /// Make an HTTP request to a given URL and return the HTTP response, possibly after a transformation.
 #[cfg(target_arch = "wasm32")]
 pub async fn http_request(arg: CanisterHttpRequestArgument) -> CallResult<(HttpResponse,)> {
-    ic_cdk::api::management_canister::http_request::http_request(arg).await
+    ic_cdk::api::management_canister::http_request::http_request_with_cycles(arg, 0).await
```
ielashi, line 18: Isn't this crate meant to be general-purpose? If yes, then this won't work for anyone using this library in canister that's running on an app subnet.

maksymar: Right, fixed, thank you!

ielashi (approved):

PR #200, opened 2023-04-28
Update uploader documentation

pakhomov-dfinity:


bootstrap/README.md
```diff
@@ -1,4 +1,4 @@
-# Computing the Bitcoin State
+# Uploading the Bitcoin State
```
ielashi, line 1:
```suggestion
# Bootstrapping the Bitcoin State
```
I think bootstrapping is the more accurate term, as bootstrapping = computing + uploading

bootstrap/README.md
```diff
@@ -27,7 +27,14 @@ STABILITY_THRESHOLD=<desired stability threshold>
 NETWORK=<mainnet or testnet>
 ```
 
-## 3. Download the Bitcoin state
+## 3. Checkout the working revision
+See [Proposal](https://dashboard.internetcomputer.org/proposal/94253)
+
+```
+git checkout fdfced51d002d7f16908642f29216a869ecd3627
+```
```
ielashi, line 35: The steps here are not specific to computing the state we did in this proposal, so I'd remove this. To validate that specific proposal, you should follow the instructions in the [launch branch](https://github.com/dfinity/bitcoin-canister/tree/launch/bootstrap).

bootstrap/README.md
```diff
@@ -77,3 +84,59 @@ cargo run --release --example compute_hashes -- --file ./canister_state.bin > ch
...
+cp bootstrap/chunk_hashes.txt bootstrap/uploader/src/chunk_hashes.txt
+```
+
+Update the `state_size` in `bootstrap/uploader/src/main.rs` to be equal to the state size in 64k blocks, i.e. `echo $(( $(stat -c %s bootstrap/canister_state.bin) / 65536 ))`
+
+## 8. Deploy the uploader canister
+
+`dfx` version `0.12.1` is known to work
+
+Add desired network to the `dfx.json` and set `NETWORK` to match it.
```
ielashi, line 100: The dfx.json is already at version 0.13.1. Why go back to 0.12.1?

bootstrap/README.md
```diff
@@ -77,3 +84,59 @@ cargo run --release --example compute_hashes -- --file ./canister_state.bin > ch
...
+```
+cd uploader
+cargo run --example upload -- --canister-id $CANISTER --state ../canister_state.bin --ic-network http://[${IP}]:8080  --fetch-root-key
+```
+
+If the desired node gets unhealthy, the uploader crashes but it continues to upload on restart. Takes few hours to few days.
+
+## 10. Hardcode the canister id into the replica binary
+
+Set the bitcoin canister id in `rs/config/src/execution_environment.rs` to match the uploader canister id. Note that "mainnet" and "testnet" stand for the BTC netwrok version, not IC. Build image and upgrade the subnet.
```
ielashi, line 123: That is no longer necessary. You can now deploy a testnet with the flag `--allow-specified-ids` and with that you can create the canister with the same ID as the bitcoin canister in production.

bootstrap/README.md
```diff
@@ -77,3 +84,59 @@ cargo run --release --example compute_hashes -- --file ./canister_state.bin > ch
 ```
 
 The hash of each chunk is saved in `chunk_hashes.txt`.
+
+## 7. Hardcode the caches into the uploader canister
```
ielashi, line 88: The instructions below here are quite specific for testnets. Can we add a note perhaps that clarifies that the remaining instructions are testnet-specific?

ielashi:
Thanks the for the details instructions @pakhomov-dfinity. I left a few minor comments.

bootstrap/README.md
pakhomov-dfinity: The launch proposal has relevant updates, e.g. uploader canister accepts hardcoded list of hashes. Otherwise one needs to pass more hashes than possible to shove into dfx call as startup flags, which I failed to do.

bootstrap/README.md
pakhomov-dfinity: 0.13 doesn't create canisters on testnets

bootstrap/README.md
pakhomov-dfinity: Ack, I need to test how it works

ielashi:
We have recently updated the bootstrapping logic to address the issues mentioned in this PR, so we can close this now.

PR #203, opened 2023-05-04
feat: add watchdog stop-switch for bitcoin canister

maksymar:
This change adds reading and writing bitcoin canister API access config flag.

API access target is calculated based on the height health status, actual value is read from the bitcoin canister. If there is a difference between the target and actual values, then watchdog tries to rewrite the bitcoin canister API access config flag.

This change also exposes additional metrics: 
- blocks behind/ahead thresholds
- minimum and actual number of explorers
- API access target

### DIFF
--- a/watchdog/src/api_access.rs
+++ b/watchdog/src/api_access.rs
@@ -1,0 +1,66 @@
+use crate::health::{HealthStatus, HeightStatus};
+use ic_btc_interface::{Config as BitcoinCanisterConfig, Flag};
+
+#[derive(Clone, Debug)]
+pub struct ApiAccess {
+    /// Expected value of the Bitcoin canister API access flag.
+    pub target: Option<Flag>,
+
+    /// Actual value of the Bitcoin canister API access flag.
+    pub actual: Option<Flag>,
+}
+
+impl ApiAccess {
+    pub fn new() -> Self {
+        Self {
+            target: None,
+            actual: None,
+        }
+    }
+}
+
+impl Default for ApiAccess {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+/// Calculates the target value of the Bitcoin canister API access flag.
+fn calculate_target(health: HealthStatus) -> Option<Flag> {
+    match health.height_status {
+        HeightStatus::Ok => Some(Flag::Enabled),
+        HeightStatus::Behind | HeightStatus::Ahead => Some(Flag::Disabled),
+        HeightStatus::NotEnoughData => None,
+    }
+}
+
+async fn get_bitcoin_canister_config() -> Option<BitcoinCanisterConfig> {
+    // TODO: read the configuration from the Bitcoin canister.
+    Some(BitcoinCanisterConfig::default())
+}
+
+/// Fetches the API access flag from the Bitcoin canister.
+pub async fn fetch_api_access() {
+    let health = crate::health::health_status();
+    let target = crate::api_access::calculate_target(health);
+
+    let bitcoin_canister_config = get_bitcoin_canister_config().await;
+    let actual = bitcoin_canister_config.map(|config| config.api_access);
+
+    let api_access = ApiAccess { target, actual };
+
+    crate::storage::set_api_access(api_access);
+}
+
+/// Sets the API access flag in the Bitcoin canister.
+pub async fn set_api_access() {
+    let api_access = crate::storage::get_api_access();
+    match (api_access.target, api_access.actual) {
+        (None, _) => (),
+        (Some(target), actual) => {
+            if Some(target) != actual {
+                // TODO: set the API access in the Bitcoin canister.
+            }
+        }
+    }
+}
--- a/watchdog/src/api_access.rs
+++ b/watchdog/src/api_access.rs
@@ -4,7 +4,6 @@ use candid::CandidType;
 use ic_btc_interface::{Config as BitcoinCanisterConfig, Flag, SetConfigRequest};
 
 /// Captures the expected and the actual value of the Bitcoin canister API access flag.
-/// They might be different if the Bitcoin canister is not reachable.
 #[derive(Clone, Debug, CandidType)]
 pub struct ApiAccess {
     /// Expected value of the Bitcoin canister API access flag.
@@ -20,6 +19,11 @@ impl ApiAccess {
             target: None,
             actual: None,
         }
+    }
+
+    /// Checks if the target and actual API access flags are in sync.
+    pub fn is_in_sync(&self) -> bool {
+        self.target == self.actual
     }
 }
 
@@ -42,19 +46,15 @@ fn calculate_target(health: HealthStatus) -> Option<Flag> {
 async fn get_bitcoin_canister_config() -> Option<BitcoinCanisterConfig> {
     let id = crate::storage::get_config().bitcoin_canister_principal;
     let result = ic_cdk::api::call::call(id, "get_config", ()).await;
-    match result {
-        Ok((config,)) => config,
-        Err(err) => {
-            print(&format!("Error getting Bitcoin canister config: {:?}", err));
-            None
-        }
-    }
+    result
+        .map(|(config,)| config)
+        .map_err(|err| print(&format!("Error getting Bitcoin canister config: {:?}", err)))
+        .ok()
 }
 
-/// Fetches the API access flag from the Bitcoin canister.
-pub async fn fetch_api_access() {
-    let health = crate::health::health_status();
-    let target = crate::api_access::calculate_target(health);
+/// Fetches the actual API access flag and calculates the target value.
+async fn fetch_actual_and_calculate_target_api_access() {
+    let target = calculate_target(crate::health::health_status());
 
     let bitcoin_canister_config = get_bitcoin_canister_config().await;
     let actual = bitcoin_canister_config.map(|config| config.api_access);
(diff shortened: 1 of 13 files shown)


watchdog/src/config.rs
```diff
@@ -78,6 +90,8 @@ impl Config {
             blocks_behind_threshold: BLOCKS_BEHIND_THRESHOLD,
             blocks_ahead_threshold: BLOCKS_AHEAD_THRESHOLD,
             min_explorers: MIN_EXPLORERS,
+            bitcoin_canister_principal: Principal::from_text(MAINNET_BITCOIN_CANISTER_PRINCIPAL)
```
ielashi, line 93: Shouldn't the endpoint below be a function of the canister principal?

watchdog/src/config.rs
```diff
@@ -91,6 +105,8 @@ impl Config {
             blocks_behind_threshold: BLOCKS_BEHIND_THRESHOLD,
             blocks_ahead_threshold: BLOCKS_AHEAD_THRESHOLD,
             min_explorers: MIN_EXPLORERS,
+            bitcoin_canister_principal: Principal::from_text(TESTNET_BITCOIN_CANISTER_PRINCIPAL)
```
ielashi, line 108: ditto

watchdog/src/metrics.rs
```diff
@@ -33,18 +36,41 @@ pub fn get_metrics() -> CandidHttpResponse {
...
         BitcoinNetwork::Mainnet => (1.0, 0.0),
         BitcoinNetwork::Testnet => (0.0, 1.0),
     };
     w.gauge_vec("bitcoin_network", "Bitcoin network.")?
         .value(&[("network", "mainnet")], mainnet)?
         .value(&[("network", "testnet")], testnet)?;
+    w.encode_gauge(
+        "blocks_behind_threshold",
+        // blocks_behind_threshold must be converted to negative.
+        -1.0 * config.blocks_behind_threshold as f64,
```
ielashi, line 61: I don't understand why this comment. Can you elaborate?

watchdog/src/metrics.rs
```diff
@@ -75,19 +101,45 @@ fn encode_metrics(w: &mut MetricsEncoder<Vec<u8>>) -> std::io::Result<()> {
...
+        if !height.is_nan() {
+            available_explorers_count += 1;
+        }
         gauge = gauge.value(&[("explorer", &explorer.to_string())], height)?;
     }
+    w.encode_gauge(
+        "available_explorers",
+        available_explorers_count as f64,
+        "The number of available explorers to compare against.",
+    )?;
```
ielashi, line 142: If I understand this correctly, an explorer would be marked as available as long as it responded at least once in the past, but if it has, say, been offline or unresponsive for several days, it would still be considered available by this logic, correct? Is that the intent? Would we still want to consider these explorers available in this case?

watchdog/src/api_access.rs
```diff
@@ -0,0 +1,102 @@
...
+        Err(err) => {
+            print(&format!("Error setting Bitcoin canister config: {:?}", err));
+        }
+    }
+}
+
+/// Synchronizes the API access flag of the Bitcoin canister, attempting a limited number of times.
+pub async fn synchronise_api_access() {
+    const ATTEMPTS: u8 = 3;
+    for _ in 0..ATTEMPTS {
```
ielashi, line 88: nit: I don't think we actually need to attempt multiple times, as there's no reason why this wouldn't work in a single pass.

watchdog/src/lib.rs
```diff
@@ -28,6 +30,9 @@ thread_local! {
     /// The local storage for the data fetched from the external APIs.
     static BLOCK_INFO_DATA: RwLock<HashMap<BitcoinBlockApi, BlockInfo>> = RwLock::new(HashMap::new());
 
+    /// The local storage for the API access.
+    static API_ACCESS: RwLock<ApiAccess> = RwLock::new(ApiAccess::new());
```
ielashi, line 34: nit: I guess this applies to other fields here as well. Why do we use a `RwLock`? I think a `RefCell` would be closer to what you need. See https://mmapped.blog/posts/01-effective-rust-canisters.html#use-threadlocal

watchdog/src/config.rs
maksymar: Initially I thought of that too. But I suppose that `<principal>.raw.ic0.app` is also a temporary value and may change some time in the future. Therefore the simplest solution I found is to have two constants with principal and a test to verify that principals are the same.

watchdog/src/lib.rs
```diff
@@ -74,6 +85,12 @@ pub fn get_config() -> Config {
     crate::storage::get_config()
 }
 
+/// Returns the API access of the Bitcoin canister.
```
dsarlis, line 88:
```suggestion
/// Returns the API access of the Bitcoin canister as known by the watchdog.
```

basically make it clear that this is the local information about the api access which might or might not be in sync.

watchdog/src/metrics.rs
maksymar: There are 2 threshold values for the height_diff: blocks_behind_threshold and blocks_ahead_threshold. One must be always negative, another is always positive, eg. [-3, 5]. But if I use say `i64` type, then it'd be technically possible to provide an invalid config, eg. [3, -5]. To avoid that I decided to always store only values, but add a sign when the value is used.

Maybe a better option would be to add an extra function that would handle the sign only in one place.

watchdog/src/api_access.rs
```diff
@@ -0,0 +1,102 @@
...
+    match result {
+        Ok(()) => (),
+        Err(err) => {
+            print(&format!("Error setting Bitcoin canister config: {:?}", err));
+        }
+    }
+}
+
+/// Synchronizes the API access flag of the Bitcoin canister, attempting a limited number of times.
+pub async fn synchronise_api_access() {
```
dsarlis, line 86: The one point I'm not entirely sure about is whether we want to immediately turn on the flag again if we believe that the canister is again in sync with explorers.

I think disabling the access is totally fine (that's the goal of this work) but enabling automatically without someone checking why we had to disable in the first place and whether there is a deeper issue with the bitcoin canister is not so clear imo. @ielashi wdyt?

dsarlis (commented):
Do we have a test to check whether the config can be changed properly? Probably we need an end-to-end test for that?

watchdog/src/metrics.rs
maksymar: Switched to a getter function. PTAL.

watchdog/src/metrics.rs
maksymar: No, it's not exactly right. Currently the information from explorers is stored till the next fetch, so if explorer failed to provide data it will be erased. That's why the number of explorers can vary from fetch to fetch (every 2 minutes currently), and that's why we have `min_explorers` threshold of 2 with 6 explorers polled.

watchdog/src/api_access.rs
maksymar: I agree, but this approach gives a bit better metrics charts.

Since fetching explorers data and API access synchronisation happens periodically every 2 minutes and that the metrics are populated from the data stored in storage it's important to (1) fetch data and calculate target api access value (2) send update call to bitcoin canister and (3) read bitcoin config to validate api access flag.

it may be done sequentially or in a loop. I chose to do it in a loop where 2 attempts would be enough and steps (1)-(2) will be covered on the first iteration and step (3) at the beginning of the second iteration. I put `ATTEMPS = 3` just in case, does not hurt.

am I over complicating this? :)

watchdog/src/config.rs
maksymar: Done.

watchdog/src/config.rs
maksymar: Done.

watchdog/src/lib.rs
maksymar: Thank you, done!

watchdog/src/api_access.rs
maksymar: UPD: after an off-line discussion with Dimitris I removed `ApiAccess` structure and the loop-approach. Now it's only storing the target flag and updates the API access if the target is present and is different from the actual Bitcoin canister value.

watchdog/src/lib.rs
maksymar: Not relevant any more after removing the `ApiAccess` struct. Now it's only storing the API access target value.

watchdog/src/api_access.rs
ielashi: Do you see a particular danger in turning it on again should we be synced again? As I think about, I don't see how that can be dangerous.

There's a bit of an annoyance with the current approach though, and that is that the watchdog is a bit too persistent in updating the config. For example, what if we wanted to manually disable the API even though we're synced because we discovered something else being broken? In that case, the watchdog canister would immediately enable the API again.

Maybe the watchdog should be more gentle, and only attempt to update the config once should the `target` change, and not any other time.

maksymar: Another idea: add a latch to the bitcoin canister that on every manual API turn off would automatically drop watchdog permissions for changing the API access.

So the maintenance steps would be:
- manually disable bitcoin canister API (automatically removes watchdog permissions)
- update bitcoin canister
- manually enable bitcoin canister API access
- (manually or automatically with the previous step) grant permissions for watchdog

watchdog/src/config.rs
```diff
@@ -44,16 +44,17 @@ pub struct Config {
     pub bitcoin_network: BitcoinNetwork,
 
     /// Below this threshold, the canister is considered to be behind.
-    pub blocks_behind_threshold: u64,
+    /// This value is positive, but it must be converted to negative on every usage.
+    blocks_behind_threshold: u64,
```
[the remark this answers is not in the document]
maksymar: Removed the comment about the "conversion".

watchdog/src/api_access.rs
ielashi: > Alternatively we can add a turn on/off button for a watchdog canister. So for every bitcoin canister it would require to run the watchdog off, do the thing on bitcoin canister and then turn the watchdog canister on again.

Right, that 's actually already possible with our current configuration. Whenever we want to manually override the watchdog, we can also set the `watchdog_canister` in the bitcoin canister's configuration to be `none`, and then it won't listen to the watchdog commands, and we can reset it afterwards. I'll add that to the instructions in the runbook.

dsarlis: > Do you see a particular danger in turning it on again should we be synced again? As I think about, I don't see how that can be dangerous.

Not necessarily dangerous but it could also lead to some turn off/turn on loop which would be also annoying. I think it's mostly about making sure that we look into cases where the bitcoin canister fell behind and understand what happened as opposed to just ignore because after a few minutes it was again syncing and then we re-enabled access. I suppose if we get notified whenever the watchdog needs to intervene that would also be ok, we can look into issues/improvements afterwards.

> For example, what if we wanted to manually disable the API even though we're synced because we discovered something else being broken? In that case, the watchdog canister would immediately enable the API again.

Yes, that's another good point. As discussed, we can, in that case, remove the permission for `watchdog_canister` and manually disable the apis.

maksymar: To conclude the discussion, I suppose, we agreed that the bitcoin canister should handle watchdog's permissions, either manually or automatically. Since there is no action points for the watchdog I'm marking this conversation resolved. Feel free to reopen.

maksymar:
LGTMs?

ielashi (approved):

PR #206, opened 2023-05-11
chore: prepare watchdog for e2e tests

maksymar:


watchdog/e2e-tests/get_config.sh
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env bash
```
dsarlis, line 1: Shouldn't we rename these with a `git mv` to preserve history? Now it will look like a new file. Ditto for the other two files.

dsarlis (approved):

PR #207, opened 2023-05-15
chore: expose main_chain_height endpoint for watchdog

maksymar:
This change exposes `main_chain_height` for the watchdog.

The goal is to switch from reading it via metrics page parsing to making an inter-canister call.


ielashi (changes requested):
**Response on slack:**

> Hey Maksym, I’d hold off on that change for a bit, as that requires a change in the internet computer specification.

> Other people also requested a similar endpoint, but we’ll have to think a bit more in terms of how we want to expose it, and what other information we should include as well.

PR #209, opened 2023-06-05
feat: add configurable explorers for watchdog canister

maksymar:
This PR adds configurable bitcoin explorers to the watchdog canister.

Additionally
- change the fetching duration from 2 to 4 minutes (not to abuse explorer APIs)
- add deployment script to simplify deploying watchdog canister to mainnet/testnet
- updates some tests due to the changed `providers` order (bitcoin canister now comes at the end)

### DIFF
--- a/watchdog/scripts/deploy.sh
+++ b/watchdog/scripts/deploy.sh
@@ -1,0 +1,72 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# Hardcoded values.
+BITCOIN_CANISTER_ID_MAINNET=ghsi2-tqaaa-aaaan-aaaca-cai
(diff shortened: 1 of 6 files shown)


watchdog/scripts/deploy.sh
```diff
@@ -0,0 +1,77 @@
+#!/usr/bin/env bash
```
dsarlis, line 1: I don't think this script would be useful really in production. I expect that we'll assign controllership of the watchdog to the NNS and it would then be only upgradable through proposals.

maksymar: Done, removed.

dsarlis (approved):

PR #210, opened 2023-06-08
chore: add watchdog-canister to docker

maksymar:
This PR adds a watchdog canister to a docker file to build and verify its integrity.

Additionally:
- added reproducibility CI check, which runs docker build 2 times and compares the check sums
- added top-level comment to Dockerfile
- minimised the number of layers by combining commands in a single `RUN` with `&&`
- added packet manager cache cleanup to reduce the image size

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -447,6 +447,17 @@ jobs:
         run: |
           bash watchdog/e2e-tests/metrics.sh
 
+  reproducibility:
+    runs-on: ubuntu-20.04
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+
+      - name: Verify Reproducibility
+        run: |
+          ./scripts/verify_reproducibility.sh Dockerfile
+
   checks-pass:
     needs:
       - cargo-tests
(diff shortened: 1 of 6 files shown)


.github/workflows/workflow.yml
```diff
@@ -447,6 +447,17 @@ jobs:
         run: |
           bash watchdog/e2e-tests/metrics.sh
 
+  reproducibility:
```
dsarlis, line 450: Should we name this something a bit more descriptive, like `canister-build-reproducibility`?

e2e-tests/reproducibility.sh
```diff
@@ -0,0 +1,59 @@
...
+# Calculate the SHA256 sums for the second build
+echo "Calculating SHA256 sums (2nd build)..."
+sha256sum2=$(sha256sum "$tmpdir/watchdog-canister.wasm.gz" "$tmpdir/uploader-canister.wasm.gz" "$tmpdir/ic-btc-canister.wasm.gz")
+
+# Compare the SHA256 sums
+if [ "$sha256sum1" = "$sha256sum2" ]; then
+  echo "SUCCESS: Reproducible build, SHA256 sums match."
+  exit 0
+else
+  echo "FAIL: Non-reproducible build, SHA256 sums differ."
```
dsarlis, line 57: Should we also print the check sums for easy inspection of what's different? E.g. maybe it's just one canister not all of them.

dsarlis (approved):

e2e-tests/reproducibility.sh
maksymar: Good point, done.

PR #211, opened 2023-06-09
chore: add watchdog_canister to set_config_request api

maksymar:
This PR adds the ability to set a watchdog canister principal in a bitcoin canister config via `set_config` API, which is used to update the bitcoin canister config via proposals.

The watchdog_canister principal ID can be flexibly configured according to your needs. You can either assign an actual principal ID or reset it back to None, providing more control and customization for your canister settings.

### DIFF
--- a/canister/candid.did
+++ b/canister/candid.did
@@ -88,7 +88,7 @@ type set_config_request = record {
   fees : opt fees;
   api_access : opt flag;
   disable_api_if_not_fully_synced : opt flag;
-  watchdog_canister : opt principal;
+  watchdog_canister : opt opt principal;
 };
 
 service bitcoin : (config) -> {
--- a/canister/candid.did
+++ b/canister/candid.did
@@ -88,6 +88,7 @@ type set_config_request = record {
   fees : opt fees;
   api_access : opt flag;
   disable_api_if_not_fully_synced : opt flag;
+  watchdog_canister : opt principal;
 };
 
 service bitcoin : (config) -> {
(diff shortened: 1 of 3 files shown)


canister/candid.did
```diff
@@ -88,6 +88,7 @@ type set_config_request = record {
   fees : opt fees;
   api_access : opt flag;
   disable_api_if_not_fully_synced : opt flag;
+  watchdog_canister : opt principal;
```
dsarlis, line 91: Shouldn't this be `opt opt principal`? Otherwise, it's not matching the Rust type.

canister/src/api/set_config.rs
```diff
@@ -55,6 +55,9 @@ fn set_config_no_verification(request: SetConfigRequest) {
         if let Some(disable_api_if_not_fully_synced) = request.disable_api_if_not_fully_synced {
             s.disable_api_if_not_fully_synced = disable_api_if_not_fully_synced;
         }
+        if let Some(watchdog_canister) = request.watchdog_canister {
+            s.watchdog_canister = watchdog_canister;
+        }
```
dsarlis, line 60: We should also probably add a test for this case, similar to other remaining cases setting a field of the config.

canister/candid.did
maksymar: Right, thank you! Done.

canister/src/api/set_config.rs
maksymar: Done.

dsarlis (approved):

PR #213, opened 2023-06-13
chore: upgrade candid version to 0.9.0-beta.3

dragoljub-djuric:


maksymar (approved):

ielashi:
Curious, why are we upgrading to a beta version of candid?

dragoljub-duric:
@ielashi We need a new candid version to safely add the new 'evict' install canister mode to the replica. Hence upgrading the candid version on the replica required an upgrade on the Bitcoin canister side.

PR #216, opened 2023-06-15
chore: replace CDK http_request_with_cycles with direct call_with_payment128

maksymar:
This PR replaces the `http_request_with_cycles` function call from `ic_cdk` with its underlying `call_with_payment128`.

It is done to avoid dependency issues in the recent repo updates.

### DIFF
--- a/ic-http/src/http_request.rs
+++ b/ic-http/src/http_request.rs
@@ -1,11 +1,6 @@
 use ic_cdk::api::call::RejectionCode;
 use ic_cdk::api::management_canister::http_request::{CanisterHttpRequestArgument, HttpResponse};
 
-#[cfg(target_arch = "wasm32")]
-use candid::Principal;
-#[cfg(target_arch = "wasm32")]
-use ic_cdk::api::call::call_with_payment128;
-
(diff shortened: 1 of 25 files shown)


ic-http/src/http_request.rs
```diff
@@ -34,5 +39,14 @@ pub async fn http_request_with_cycles(
     arg: CanisterHttpRequestArgument,
     cycles: u128,
 ) -> CallResult<(HttpResponse,)> {
-    ic_cdk::api::management_canister::http_request::http_request_with_cycles(arg, cycles).await
+    // NOTE: removing `http_request_with_cycles` is a workaround for the dependency issue
+    // when a recent `ic_cdk` version lost the `http_request_with_cycles` function.
+    // TODO: put back `http_request_with_cycles` as soon as it's available in the corresponding CDK version.
```
dsarlis, line 44: Not sure it's really worth leaving a TODO. The current version looks clean enough to me that I would actually keep it forever, the wrapper hardly adds value imo.

dsarlis (approved):

PR #220, opened 2023-06-30
refactor: introduce Txid type

roman-kashitsyn:
This change improves the type safety of the Bitcoin interface by introducing a newtype for Bitcoin transaction identifiers.

### DIFF
--- a/interface/src/lib.rs
+++ b/interface/src/lib.rs
@@ -139,6 +139,32 @@ impl<'de> serde::de::Deserialize<'de> for Txid {
                     Err(_) => Err(E::invalid_length(value.len(), &self)),
                 }
             }
+
+            fn visit_seq<A>(self, mut seq: A) -> Result<Self::Value, A::Error>
+            where
+                A: serde::de::SeqAccess<'de>,
+            {
+                use serde::de::Error;
+                if let Some(size_hint) = seq.size_hint() {
+                    if size_hint != 32 {
+                        return Err(A::Error::invalid_length(size_hint, &self));
+                    }
+                }
+                let mut bytes = [0u8; 32];
+                let mut i = 0;
+                while let Some(byte) = seq.next_element()? {
+                    if i == 32 {
+                        return Err(A::Error::invalid_length(i + 1, &self));
+                    }
+
+                    bytes[i] = byte;
+                    i += 1;
+                }
+                if i != 32 {
+                    return Err(A::Error::invalid_length(i, &self));
+                }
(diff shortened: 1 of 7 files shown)


interface/src/lib.rs
```diff
@@ -97,15 +98,133 @@ impl std::fmt::Display for NetworkInRequest {
...
+                match TryInto::<[u8; 32]>::try_into(value) {
+                    Ok(txid) => Ok(Txid(txid)),
+                    Err(_) => Err(E::invalid_length(value.len(), &self)),
+                }
+            }
+        }
+
+        deserializer.deserialize_bytes(TxidVisitor)
+    }
+}
```
dsarlis: Can you not avoid this by having the `#[serde_bytes(with = "serde_bytes")]` annotation in the type definition?

interface/src/lib.rs
```diff
@@ -97,15 +98,133 @@ impl std::fmt::Display for NetworkInRequest {
...
+
+impl fmt::Display for Txid {
+    fn fmt(&self, fmt: &mut fmt::Formatter<'_>) -> fmt::Result {
+        // In Bitcoin, you display hash bytes in reverse order.
+        //
+        // > Due to historical accident, the tx and block hashes that bitcoin core
+        // > uses are byte-reversed. I’m not entirely sure why. Maybe something
+        // > like using openssl bignum to store hashes or something like that,
+        // > then printing them as a number.
+        // > -- Wladimir van der Laan
```
dsarlis, line 182: Lol 🤣

interface/src/lib.rs
dsarlis: I guess you want to be able to return custom errors when deserialization fails but then maybe we use `serde(deserialize_with)` and pass a function that does the checks you want.

interface/src/lib.rs
THLO: It's a hobby project. What did you expect?   :slightly_smiling_face:

interface/src/lib.rs
roman-kashitsyn: serde_bytes cannot handle byte arrays.

dsarlis: Oh no... Ok, then.

dsarlis (approved):

interface/src/lib.rs
roman-kashitsyn: I also considered https://docs.rs/serde_arrays/latest/serde_arrays/, but these can't handle blobs.

PR #221, opened 2023-06-30
fix: fix finding next difficulty in chain for testnets

maksymar:
This PR fixes finding the next difficulty in chain for testnets when verifying headers.

Previous implementation was jumping over one block hash because it was advancing the current header and then was saving its previous block hash (which turned to be a prev-prev hash, one-off bug).
Instead it should have saved the previous block hash first (to avoid its calculation on every iteration) and only then advancing the current header. Or another option for readability to recalculate hash after advancing to the previous header.

Previous implementation:
```rust
while current_hash != initial_header_hash {
    // It does not check initial header for PoW.
    if is_pow_found {
        return current_header.bits;
    }

    // Advance current header.
    current_header = store
        .get_with_block_hash(&current_header.prev_blockhash)
        .expect("previous header should be in the header store");
    // Attempt to save the prev hash, but it's now actually prev-prev-hash.
    current_hash = current_header.prev_blockhash;
}
pow_limit_bits
```

New implementation:
```rust
loop {
    // First check if PoW found...
    if is_pow_found {
        return current_header.bits;
    }
    // ...and only then stop when it's the initial header.
    if current_hash == initial_header_hash {
        break;
    }

    // Advance to the previous header.
    current_header = store
        .get_with_block_hash(&current_header.prev_blockhash)
        .expect("previous header should be in the header store");
    // Update current hash.
    current_hash = current_header.block_hash();
}
pow_limit_bits
```

### DIFF
--- a/validation/src/header.rs
+++ b/validation/src/header.rs
@@ -198,7 +198,7 @@ fn get_next_target(
 /// returns to its previous value." This function is used to compute the
 /// difficulty target in case the block has been found within 20
 /// minutes.
-fn find_next_difficulty_in_chain(
+fn find_next_difficulty_in_chain0(
     network: &Network,
     store: &impl HeaderStore,
     prev_header: &BlockHeader,
@@ -229,6 +229,52 @@ fn find_next_difficulty_in_chain(
                 current_height -= 1;
                 current_hash = current_header.prev_blockhash;
             }
+            pow_limit_bits
+        }
+        Network::Bitcoin | Network::Signet => pow_limit_bits,
+    }
+}
+
+/// This method is only valid when used for testnet and regtest networks.
+/// As per "https://en.bitcoin.it/wiki/Testnet",
+/// "If no block has been found in 20 minutes, the difficulty automatically
+/// resets back to the minimum for a single block, after which it
+/// returns to its previous value." This function is used to compute the
+/// difficulty target in case the block has been found within 20
+/// minutes.
+fn find_next_difficulty_in_chain(
+    network: &Network,
+    store: &impl HeaderStore,
+    header: &BlockHeader,
+    height: BlockHeight,
+) -> u32 {
+    // This is the maximum difficulty target for the network
+    let pow_limit_bits = pow_limit_bits(network);
+    match network {
+        Network::Testnet | Network::Regtest => {
+            let mut current_header = *header;
+            let mut current_height = height;
+            let initial_header_hash = store.get_initial_hash();
+
+            // Keep traversing the blockchain backwards from the recent block to initial
+            // header hash.
+            loop {
+                if current_header.bits != pow_limit_bits
+                    || current_height % DIFFICULTY_ADJUSTMENT_INTERVAL == 0
+                {
+                    return current_header.bits;
+                }
+                if current_header.block_hash() == initial_header_hash {
+                    break;
+                }
+
+                // Traverse to the previous header
+                current_header = store
+                    .get_with_block_hash(&current_header.prev_blockhash)
+                    .expect("previous header should be in the header store");
+                current_height -= 1;
+            }
+
             pow_limit_bits
         }
         Network::Bitcoin | Network::Signet => pow_limit_bits,
@@ -663,7 +709,7 @@ mod test {
             find_next_difficulty_in_chain(&network, &store, &last_header, chain_length - 1);
 
         // Assert.
-        assert_eq!(result, pow_limit_bits(&network));
+        assert_eq!(result, FAKE_POW);
     }
 
     #[test]
--- a/validation/src/header.rs
+++ b/validation/src/header.rs
@@ -668,7 +668,7 @@ mod test {
     #[test]
     fn test_find_next_difficulty_in_chain_pow_found() {
         // This test checks the chain of headers of different lengths
-        // with non-limit PoW in the first block header and Pow limit
+        // with non-limit PoW in the first block header and PoW limit
         // in all the other headers.
         // Expect difficulty to be equal to the non-limit PoW.
         // Expect headers inspected to be equal to the chain length.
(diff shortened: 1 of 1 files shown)


validation/src/header.rs
```diff
@@ -203,35 +200,46 @@ fn find_next_difficulty_in_chain(
...
+                }
+
+                // Stop if we reach the initial header.
+                if current_hash == initial_header_hash {
+                    break;
                 }
 
-                // Traverse to the previous header
+                // Save the previous header's hash, so that there's no need to calculate it.
+                current_hash = current_header.prev_blockhash;
```
ielashi, line 233: nit: given that this is code that's only relevant for testnet/regtest, I wouldn't worry too much about optimizations and would prioritize readability instead. It seems that the only change we really need is changing line 230 in the previous version of the file from:

```
current_hash = current_header.prev_blockhash;
```

to:

```
current_hash = current_header.block_hash();
```

I'd personally find that logic more straightforward to read.

validation/src/header.rs
```diff
@@ -203,35 +200,46 @@ fn find_next_difficulty_in_chain(
     store: &impl HeaderStore,
     prev_header: &BlockHeader,
     prev_height: BlockHeight,
-) -> u32 {
+) -> (u32, u32) {
```
ielashi, line 203: IIUC, here we're exposing some internal information, the number of headers inspected, to verify it in our tests. I wonder if this is necessary. Here's my reasoning:

1. If it's necessary, then that means that our coverage of this method is insufficient, since we should be able to come up with examples where the difficulty is wrong if the number of headers is wrong.
2. If it's not necessary, then we have enough coverage for this method, so why have it at all?

validation/src/header.rs
maksymar: Done, switched to using `block_hash()` for readability.

maksymar: Note: there are 2 changes
1) using `current_header.block_hash()` instead of `current_header.prev_blockhash` in order not to jump over one block
2) replacing while-condition to check the difficulty of the initial block and only then to stop.

But this (2) change breaks `test_is_header_valid_invalid_computed_target` test, because its  initial `header_705600` is actually having non-limit PoW that breaks an expectation of `ValidateHeaderError::InvalidPoWForComputedTarget`.

If my assumption about checking all the headers (including the initial one) is correct, then this `test_is_header_valid_invalid_computed_target` has to be modified. One option to modify it would be to have blocks with PoW limit, so it finds no target and fails with an expected error. WDYT?

maksymar: I've modified `test_is_header_valid_invalid_computed_target` to make it pass. PTAL.

validation/src/header.rs
maksymar: Removed `headers_inspected` counter.

validation/src/header.rs
```diff
@@ -607,4 +620,97 @@ mod test {
...
+        }
+
+        (store, last_header)
+    }
+
+    const SOME_NON_LIMIT_POW: u32 = 7;
+    const CHAIN_LENGTH_MAX: u32 = 10;
+
+    #[test]
+    fn test_find_next_difficulty_in_chain_pow_found() {
```
ielashi, line 669: I'm not sure these tests are adding coverage, since they pass without the bug fixes of this PR. I think we should have at least one test case that fails before your change, and now passes with your change.

maksymar: I agree that `test_find_next_difficulty_in_chain_pow_not_found` do not fail without those changes. But the ones you pointed out `test_find_next_difficulty_in_chain_pow_found` do fail without those changes.

I removed `test_find_next_difficulty_in_chain_pow_not_found`.

validation/src/header.rs
```diff
@@ -607,4 +620,72 @@ mod test {
...
+            assert_eq!(store.height() + 1, chain_length);
+
+            // Act.
+            let difficulty =
+                find_next_difficulty_in_chain(&network, &store, &last_header, chain_length - 1);
+
+            // Assert.
+            assert_eq!(difficulty, expected_pow);
+        }
+    }
```
ielashi, line 690: I think this test should be at the level of `get_next_target` and not calling `find_next_difficulty_in_chain`. The latter is a small helper method and doesn't capture the big picture. It's also poorly named, but that's unrelated to this change.

```suggestion
    #[test]
    fn test_next_target_regtest() {
        // This test checks the chain of headers of different lengths
        // with non-limit PoW in the first block header and PoW limit
        // in all the other headers.
        // Expect difficulty to be equal to the non-limit PoW.

        // Arrange.
        let network = Network::Regtest;
        let expected_pow = SOME_NON_LIMIT_POW;

        for chain_length in 1..CHAIN_LENGTH_MAX {
            let (store, last_header) = create_chain(&network, expected_pow, chain_length);
            assert_eq!(store.height() + 1, chain_length);

            // Act.
            let target = get_next_target(
                &network,
                &store,
                &last_header,
                chain_length - 1,
                &next_block_header(last_header, expected_pow)
            );

            // Assert.
            assert_eq!(target, BlockHeader::u256_from_compact_target(expected_pow));
        }
    }

```

ielashi (approved):

validation/src/header.rs
```diff
@@ -607,4 +620,72 @@ mod test {
...
+            let new_header = next_block_header(last_header, pow_limit);
+            store.add(new_header);
+            last_header = new_header;
+        }
+
+        (store, last_header)
+    }
+
+    const SOME_NON_LIMIT_POW: u32 = 7;
+    const CHAIN_LENGTH_MAX: u32 = 10;
```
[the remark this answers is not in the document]
maksymar: Done.

validation/src/header.rs
maksymar: Done.

PR #222, opened 2023-07-04
fix: make api_access metric an enum

maksymar:
This PR changes the `api_access` metric to an enum with clear `enabled` and `disabled` states for both Bitcoin and watchdog canisters.

Previously, we used `0/1` to represent `disabled/enabled`, but Grafana views no value as zero, causing confusion on charts.

Now, with enums, dashboards will clearly show `enabled` or `disabled`, and won't assume a state if there's no value. 

This is a "nice to have" improvement.


ielashi (approved):
LGTM.

PR #223, opened 2023-07-05
test: add proptests for computing next header target.

ielashi:
The `get_next_target` doesn't have good test coverage, yet it's important and a bug in it can bring the bitcoin integration to a halt or, in a worse case, provide a window for an attacker to feed bad blocks into the system.

This PR adds proptests to add more test coverage to this code.


dsarlis (approved):
Lgtm. Clippy seems to have some opinions though...

PR #225, opened 2023-07-07
fix: Use vbyte for the computation of a transaction fee

AlexandraZapuc:
The specification mentions that the fee percentiles are measured in [millisatoshi/vbyte](https://github.com/dfinity/interface-spec/blob/master/spec/index.md#ic-method-bitcoin_get_current_fee_percentiles-ic-bitcoin_get_current_fee_percentiles). 
This PR updates the code in accordance with the specification.

### DIFF
--- a/canister/src/api/fee_percentiles.rs
+++ b/canister/src/api/fee_percentiles.rs
@@ -550,7 +550,6 @@ mod test {
         assert_ne!(tx.vsize(), tx.size());
         assert_eq!(tx_without_witness.vsize(), tx_without_witness.size());
 
-
         let blocks = vec![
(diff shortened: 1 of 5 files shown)


ielashi (commented):
Adding a test to ensure the correctness of the change, and to avoid future regressions, would be very useful :)

THLO:
I agree!

AlexandraZapuc:
> Adding a test to ensure the correctness of the change, and to avoid future regressions, would be very useful :)

What kind of test should I add? To test the precision?

AlexandraZapuc:
Apologies for the delay! I was ooo for a good chunk of the time. I have added a test.

ielashi (approved):
Looks great, thanks @AlexandraZapuc!

PR #234, opened 2023-07-25
chore: downgrade candid to 0.8.4

maksymar:
This PR reverts upgrading candid to 0.9.1 (#231) and updates candid from 0.8.1 to 0.8.4.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -316,8 +316,8 @@ dependencies = [
  "hex",
  "ic-btc-canister",
  "ic-btc-interface",
- "ic-cdk 0.10.0",
- "ic-cdk-macros 0.7.0",
+ "ic-cdk",
+ "ic-cdk-macros",
 ]
 
 [[package]]
@@ -465,7 +465,7 @@ dependencies = [
  "anyhow",
  "binread",
  "byteorder",
- "candid_derive 0.5.0",
+ "candid_derive",
  "codespan-reporting",
  "crc32fast",
  "data-encoding",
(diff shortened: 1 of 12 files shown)


dsarlis:
@maksymar I just noticed this. It's still a draft but was still curious, why are downgrading again? Didn't we just upgrade last week? There's been back and forth between going from candid 0.8.* to 0.9.* in the past couple of weeks which I don't understand tbh.

maksymar:
> @maksymar I just noticed this. It's still a draft but was still curious, why are downgrading again? Didn't we just upgrade last week? There's been back and forth between going from candid 0.8.* to 0.9.* in the past couple of weeks which I don't understand tbh.

Thomas made an update to `ic-btc-interface` but couldn't use it because ckBTC is still using 0.8.4.
It was not clear how much time updating IC-repo would take so I prepared this PR.
But then I talked to Leo, he is aiming to update the IC-repo by the end of the week, so we'll just wait.
And I'll keep this PR as a draft in the meantime (hopefully will close it in 1-2 weeks as not needed).

ielashi:
@maksymar Do we need this still? I'm quite sure we should _not_ be downgrading for the sake of other projects, and it should be the other projects' responsibility to upgrade candid instead.

THLO:
> @maksymar Do we need this still? I'm quite sure we should _not_ be downgrading for the sake of other projects, and it should be the other projects' responsibility to upgrade candid instead.

I agree!

PR #241, opened 2023-07-28
perf: make block header validation more efficient

ielashi:
Computing the block hash is very expensive, and resulted in an "instructions limit exceeded" on a testnet. Reducing the number of block hash calls fixes the problem.

Before this PR is ready I'd like to add some perf tests to ensure the performance on this code path doesn't regress again.


AlexandraZapuc (approved):
Lgtm! 👍

dragoljub-duric (approved):

PR #246, opened 2023-08-09
fix: add a bound on the length of the unstable chain in testnet/regtest

ielashi:
In production, we set a `stability_threshold` of ~144. However, when
we consider the stability of the chain, we consider it weighted by the
blocks' difficulties. This isn't a problem in the normal case, where
block difficulties are very similar.

There are scenarios specific to the Bitcoin testnet, however, that can
make this logic problematic. The difficulty in the Bitcoin testnet can
be reset to the minimum in case a block hasn't been found for 20 minutes.

Consider the following scenario:

* Assume a `stability_threshold` of 144.
* The anchor at height `h` has difficulty of 4642.
* The anchor will be marked as stable if the difficulty-based depth of
  the successor blocks is `stability_threshold * difficulty(anchor)` = 668,448
* The difficulty is reset to the minimum of 1.
* The canister will now need to maintain a chain of length 668,448 just to
  mark the anchor block as stable!

Very long chains can cause the shadow stacks to overflow, resulting in a
broken canister.

The pragmatic solution in this case is to bound the length of the chain. If
there's only one chain and it starts exceeding a certain length, we assume
that the anchor is stable even if the difficulty requirement hasn't been
met.

This scenario is only relevant for testnets, so this addition is safe and
has not impact on the behavior of the mainnet canister.

## Benchmarks
The `get_metrics` endpoint is now slower because of all the additional metrics we've now added,
but it's still acceptable.

```
insert_block_headers    time:   [3201.5 M Instructions 3201.5 M Instructions 3201.5 M Instructions]
                        change: [-0.0039% -0.0039% -0.0039%] (p = 0.00 < 0.05)
                        Change within noise threshold.

insert_block_headers_multiple_times
                        time:   [12.055 B Instructions 12.055 B Instructions 12.055 B Instructions]
                        change: [+0.0016% +0.0016% +0.0016%] (p = 0.00 < 0.05)
                        Change within noise threshold.

insert_300_blocks       time:   [586.23 M Instructions 586.23 M Instructions 586.23 M Instructions]
                        change: [-0.0006% -0.0006% -0.0006%] (p = 0.00 < 0.05)
                        Change within noise threshold.

get_metrics             time:   [90.709 M Instructions 90.709 M Instructions 90.709 M Instructions]
                        change: [+1019.6% +1019.6% +1019.6%] (p = 0.00 < 0.05)
                        Performance has regressed.
```

### DIFF
--- a/canister/src/api/metrics.rs
+++ b/canister/src/api/metrics.rs
@@ -79,12 +79,12 @@ fn encode_metrics(w: &mut MetricsEncoder<Vec<u8>>) -> std::io::Result<()> {
             "The number of unstable blocks.",
         )?;
         w.encode_gauge(
-            "unstable_blocks_difficulty_based_depth",
+            "unstable_blocks_depth",
             state.unstable_blocks.blocks_depth() as f64,
             "The depth of the unstable blocks.",
         )?;
         w.encode_gauge(
-            "unstable_blocks_max_depth",
+            "unstable_blocks_difficulty_based_depth",
             state.unstable_blocks.blocks_difficulty_based_depth() as f64,
             "The difficulty-based depth of the unstable blocks.",
         )?;
--- a/canister/src/api/metrics.rs
+++ b/canister/src/api/metrics.rs
@@ -35,6 +35,7 @@ pub fn get_metrics() -> HttpResponse {
 
 fn encode_metrics(w: &mut MetricsEncoder<Vec<u8>>) -> std::io::Result<()> {
     with_state(|state| {
+        // General stats
         w.encode_gauge(
             "main_chain_height",
             state::main_chain_height(state) as f64,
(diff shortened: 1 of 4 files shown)


canister/src/api/metrics.rs
```diff
@@ -56,6 +57,38 @@ fn encode_metrics(w: &mut MetricsEncoder<Vec<u8>>) -> std::io::Result<()> {
...
+            state.unstable_blocks.num_tips() as f64,
+            "The number of tips in the unstable block tree.",
+        )?;
+        w.encode_gauge(
+            "unstable_blocks_total",
+            state::get_unstable_blocks(state).len() as f64,
+            "The number of unstable blocks.",
+        )?;
+        w.encode_gauge(
+            "unstable_blocks_difficulty_based_depth",
```
dragoljub-duric, line 82: Should label here and below be changed?

canister/src/api/metrics.rs
```diff
@@ -56,6 +57,38 @@ fn encode_metrics(w: &mut MetricsEncoder<Vec<u8>>) -> std::io::Result<()> {
...
+            state.unstable_blocks.blocks_depth() as f64,
+            "The depth of the unstable blocks.",
+        )?;
+        w.encode_gauge(
+            "unstable_blocks_max_depth",
```
dragoljub-duric, line 87: And here?

canister/src/api/metrics.rs
ielashi: Good catch. Done.

maksymar (approved):

PR #253, opened 2023-09-05
perf: Add inspect_message endpoint to Bitcoin canister

dragoljub-duric:
Problem:
Replicated calls to query methods first go through consensus and then are rejected by the Bitcoin canister, hence wasting resources.

 

Solution:

Add inspect_message endpoint, which will reject all calls to the query endpoints.

### DIFF
--- a/canister/src/main.rs
+++ b/canister/src/main.rs
@@ -108,10 +108,10 @@ pub fn http_request(request: HttpRequest) -> HttpResponse {
 
 #[inspect_message]
 fn inspect_message() {
-  // Reject calls to the query endpoints as they are not supported in replicated mode.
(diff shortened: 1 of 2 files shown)


canister/src/main.rs
```diff
@@ -106,4 +106,14 @@ pub fn http_request(request: HttpRequest) -> HttpResponse {
     ic_btc_canister::http_request(request)
 }
 
+#[inspect_message]
+fn inspect_message() {
+    let inspected_maethod_name = ic_cdk::api::call::method_name();
```
ielashi, line 111:
```suggestion
    let inspected_method_name = ic_cdk::api::call::method_name();
```

ielashi (approved):

PR #256, opened 2023-10-04
chore: fix upgradability test in CI

maksymar:


ielashi (approved):
LGTM, thanks @maksymar!

PR #257, opened 2023-10-04
feat: Make cargo bench part of CI to ensure we don’t accidentally introduce performance regressions

dragoljub-duric:
Problem:
We are not checking if new PRs are introducing performance regressions.

Solution:

Add benchmark test to CI.


dsarlis (approved):

ielashi (changes requested):
I have a few comments that I'll add to the PR later today.

e2e-tests/benchmark.sh
```diff
@@ -0,0 +1,74 @@
...
+IMPROVED=$(grep -c "Performance has improved." "$LOG_FILE")
+WITHIN_NOISE=$(grep -c "Change within noise threshold." "$LOG_FILE")
+set -e
+
+# Since we have 4 benchmark tests, we expect 4 appearances of "No change in performance detected."
+# or "Performance has improved." or "Change within noise threshold." otherwise, performances are degraded.
+OCCURENCES=$((NO_CHANGE+IMPROVED+WITHIN_NOISE))
+
+if [[ $OCCURENCES != 4 ]]; then
+  echo "FAIL! Performance regressions are detected."
```
dragoljub-duric, line 70: If the current branch is not updated to contain the latest master, and it happens that there was PR that improved performances and merged, then this test may fail even though it does not introduce performance regression. But I do not see an easy way around it, maybe just in the moment of failure we can print additional notes e.g. "Make sure that branch is up to date with the latest master."

e2e-tests/benchmarking/benchmark.txt
```diff
@@ -0,0 +1,149 @@
+
```
ielashi, line 1: Why should this file be checked in?

e2e-tests/benchmark.sh
```diff
@@ -0,0 +1,74 @@
...
+NO_CHANGE=$(grep -c "No change in performance detected." "$LOG_FILE")
+IMPROVED=$(grep -c "Performance has improved." "$LOG_FILE")
+WITHIN_NOISE=$(grep -c "Change within noise threshold." "$LOG_FILE")
+set -e
+
+# Since we have 4 benchmark tests, we expect 4 appearances of "No change in performance detected."
+# or "Performance has improved." or "Change within noise threshold." otherwise, performances are degraded.
+OCCURENCES=$((NO_CHANGE+IMPROVED+WITHIN_NOISE))
+
+if [[ $OCCURENCES != 4 ]]; then
```
ielashi, line 69: This logic seems quite brittle. As this number needs to constantly be updated with the number of benchmarks that we have. And it will have to be replaced once we drop criterion and use our own benchmarking framework. Let me actually see how much work it would be to replace criterion here. It should be relatively straightforward.

e2e-tests/benchmark.sh
ielashi: I agree, I think that's acceptable.

e2e-tests/benchmarking/benchmark.txt
dragoljub-duric: It is not necessary, it can be deleted after the scrip is finished.

e2e-tests/benchmark.sh
ielashi: I got rid of criterion in favor of our benchmarking logic in #263. Once that's merged, then this logic here can be simplified:

1. No need to clone master and run `cargo bench` there.
2. We can simply run `cargo bench` on this repo and detect from the report whether or not performance has changed (searching for "improved by" and "regressed by" should be sufficient)

dragoljub-duric: Perfect, thanks.

e2e-tests/benchmark.sh
dragoljub-duric: Done.

e2e-tests/benchmark.sh
```diff
@@ -0,0 +1,54 @@
...
+export PATH="$SCRIPT_DIR/..:$PATH"
+
+ROOT_DIR="$SCRIPT_DIR/.."
+
+cd "$ROOT_DIR"
+
+# Remove downloaded didc, drun.
+trap 'rm $ROOT_DIR/didc $ROOT_DIR/drun' EXIT SIGINT
+
+get_didc_release(){
```
ielashi, line 14: I think almost all this logic belongs better in the `run-benchmark.sh` script. Otherwise, running `cargo bench` outside this script doesn't work.

e2e-tests/benchmark.sh
```diff
@@ -0,0 +1,54 @@
...
+}
+
+get_drun_release() {
+  OS=$(uname | tr '[:upper:]' '[:lower:]')
+  URL="https://github.com/dfinity/ic/releases/download/release-2023-09-27_23-01%2Bquic/drun-x86_64-${OS}.gz"
+  wget -O "drun.gz" "${URL}"
+  gzip -d drun.gz
+}
+
+get_didc_release
```
ielashi, line 36: As discussed, we should not download `didc` and `drun` every time. Instead, we can skip the download of these executables exist and have the expected checksum.

e2e-tests/benchmark.sh
dragoljub-duric: Done.

benchmarks/run-benchmark.sh
```diff
@@ -3,12 +3,47 @@
 # Runs a benchmark using drun. The script assumes drun is available on the caller's path.
 set -euo pipefail
 
+# Remove downloaded didc, drun.
+trap 'set +e && rm drun didc && set -e' EXIT SIGINT
```
ielashi, line 7: Why delete drun and didc? Don't want to cache them to continue using them in the next runs?

benchmarks/run-benchmark.sh
```diff
@@ -3,12 +3,47 @@
...
+  if [ "$OS" == "Darwin" ] && [ "$ARCH" == "arm64" ]; then
+    # Apple sillicon
+    wget -O didc https://github.com/dfinity/candid/releases/download/2023-07-25/didc-macos
+  elif [ "$OS" == "Linux" ] && [ "$ARCH" == "x86_64" ]; then
+          # Linux x86 64bit
+          wget -O didc https://github.com/dfinity/candid/releases/download/2023-07-25/didc-linux64
+      else
+          echo "Unsoported machine"
+          EXIT SIGINT
+  fi
```
ielashi, line 29: fix indentation please

benchmarks/run-benchmark.sh
```diff
@@ -3,12 +3,47 @@
...
+  OS=$(uname)
+  ARCH=$(uname -m)
+  if [ "$OS" == "Darwin" ] && [ "$ARCH" == "arm64" ]; then
+    # Apple sillicon
+    wget -O didc https://github.com/dfinity/candid/releases/download/2023-07-25/didc-macos
+  elif [ "$OS" == "Linux" ] && [ "$ARCH" == "x86_64" ]; then
+          # Linux x86 64bit
+          wget -O didc https://github.com/dfinity/candid/releases/download/2023-07-25/didc-linux64
+      else
+          echo "Unsoported machine"
```
ielashi, line 27:
```suggestion
          echo "Unsupported machine"
```

benchmarks/run-benchmark.sh
```diff
@@ -3,12 +3,47 @@
...
+
+# Add script directory to the PATH
+export PATH="$SCRIPT_DIR:$PATH"
+
+get_didc_release(){
+  OS=$(uname)
+  ARCH=$(uname -m)
+  if [ "$OS" == "Darwin" ] && [ "$ARCH" == "arm64" ]; then
+    # Apple sillicon
+    wget -O didc https://github.com/dfinity/candid/releases/download/2023-07-25/didc-macos
```
ielashi, line 22: Maybe we can move the release directory (`https://github.com/dfinity/candid/releases/download/2023-07-25`) into a constant? It's used all over the place and this facilitate changing the release later.

benchmarks/run-benchmark.sh
```diff
@@ -3,12 +3,47 @@
...
+# Remove downloaded didc, drun.
+trap 'set +e && rm drun didc && set -e' EXIT SIGINT
+
 BENCH_NAME=$1
 FILE=$(mktemp)
 
+SCRIPT_DIR="$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
+
+# Add script directory to the PATH
+export PATH="$SCRIPT_DIR:$PATH"
```
ielashi, line 15: Do we need to add drun and didc to the path? I don't think we should add drun and didc to the path, as it might conflict with other versions of drun and didc that are already on the user's path.

benchmarks/run-benchmark.sh
```diff
@@ -3,12 +3,47 @@
...
+}
+
+get_drun_release() {
+  OS=$(uname | tr '[:upper:]' '[:lower:]')
+  URL="https://github.com/dfinity/ic/releases/download/release-2023-09-27_23-01%2Bquic/drun-x86_64-${OS}.gz"
+  wget -O "drun.gz" "${URL}"
+  gzip -d drun.gz
+}
+
+if ! type "didc" > /dev/null; then
```
ielashi, line 39: Let's please verify the checksum as discussed

benchmarks/run-benchmark.sh
```diff
@@ -3,12 +3,47 @@
...
+  get_didc_release
+  chmod +x didc
+fi
+
 if ! type "drun" > /dev/null; then
```
ielashi, line 44: ditto regarding checksum

dragoljub-duric: It is done now.

benchmarks/run-benchmark.sh
dragoljub-duric: Done.

dragoljub-duric: Why do we need checksum for didc?

benchmarks/run-benchmark.sh
dragoljub-duric: It is used only in two places in the link, so I do not believe it will be beneficial.

benchmarks/run-benchmark.sh
```diff
@@ -6,11 +6,52 @@ set -euo pipefail
...
+  URL="https://github.com/dfinity/ic/releases/download/release-2023-09-27_23-01%2Bquic/drun-x86_64-${OS}.gz"
+  wget -O "drun.gz" "${URL}"
+  gzip -fd drun.gz
+  chmod +x drun
+}
+
+get_correct_drun_release() {
+  OS=$(uname | tr '[:upper:]' '[:lower:]')
+  
+  if ! type "drun" > /dev/null; then
```
ielashi, line 34: This is checking if drun is on the path. What if we had two different repositories running a test like this, and each one needed a different version of drun? This approach won't work in this case.

benchmarks/run-benchmark.sh
```diff
@@ -6,11 +6,52 @@ set -euo pipefail
...
+    install_drun
+  else 
+    DRUN_LOCATION=$(type "drun" | awk '{print $3}')
+    DRUN_SHA=$(shasum -a 256 "$DRUN_LOCATION" | awk '{ print $1 }')
+    # Check if drun exists and if the correct version is used.
+    if ! [[ "$OS" == "linux" && "$DRUN_SHA" == "7bf08d5f1c1a7cd44f62c03f8554f07aa2430eb3ae81c7c0a143a68ff52dc7f7" ]]; then
```
ielashi, line 40: Let's move these constants to the top of the file please so it's clearer whenever we want to update these versions

benchmarks/run-benchmark.sh
```diff
@@ -6,11 +6,52 @@ set -euo pipefail
 BENCH_NAME=$1
 FILE=$(mktemp)
 
-if ! type "drun" > /dev/null; then
-  echo "drun is not installed. Please add drun to your path from commit d35535c96184be039aaa31f68b48bbe45909494e."
-  exit 1
+get_didc_release(){
+  OS=$(uname)
+  ARCH=$(uname -m)
+  if [ "$OS" == "Darwin" ] && [ "$ARCH" == "arm64" ]; then
```
ielashi, line 12: I don't think this binary is arm-specific. It works on my intel mac.

benchmarks/run-benchmark.sh
dragoljub-duric: I did not understand what you mean. If you have an exact idea of the change, feel free to propose it, and I will accept it.

benchmarks/run-benchmark.sh
dragoljub-duric: In the case of drun it is obvious why we need the same release. But in the case of didc I do not see any benefit of it. I am not understanding the second part of the comment.

dragoljub-duric: Done, PTAL.

benchmarks/run-benchmark.sh
ielashi: What I mean is this:

1. If I run `type drun` on my machine, it returns `drun is /usr/local/bin/drun`, because I have downloaded `drun` before and added it to my `/usr/local/bin` folder.
2. This if statement will be false.
3. You'll check the checksums, and they'll fail, so you'll download drun
4. You'll then run the `drun` that is on my path on line 86, not the one that you downloaded.

dragoljub-duric: Since I guess we should not replace the `drun` in `/usr/local/bin`, one solution can be after we download to use `alias drun="$CURR_DIR/drun" `. But this will still fail if we have a different version of `drun` in `/usr/local/bin` and the correct one in the `$CURR_DIR`. Do you have a better idea?

ielashi: A simple way around this issue is to:

1. Check if `drun` exists in the script's folder.
2. Download `drun` to the script's folder if needed.
3. Run the benchmark using the `drun` in the local folder (`./drun ... --instruction-limit ..`)

e2e-tests/benchmark.sh
```diff
@@ -0,0 +1,16 @@
...
+
+BENCH_OUTPUT=$(cargo bench)
+
+set +e
+REGRESSIONS=$( echo "$BENCH_OUTPUT" |  grep -c "regressed by" )
+set -e
+
+if [[ $REGRESSIONS != 0 ]]; then
+  echo "FAIL! Performance regressions are detected. 
+        Please run $(cargo bench -- --persist) to update $(results.yml)"
```
ielashi, line 12: Is wrapping `cargo bench` in a `$(..)` desirable? We don't actually want to run this command as part of emitting the log. ditto for `results.yml`.

dragoljub-duric: You are right, I change it.

benchmarks/run-benchmark.sh
```diff
@@ -5,11 +5,35 @@ set -euo pipefail
...
-  exit 1
-fi
+download_drun(){
+  OS=$1
+  wget -O "drun.gz" "${DRUN_RELEASE_URL_PREFIX}${OS}.gz"
+  gzip -fd drun.gz
+  chmod +x drun
+}
+
+get_correct_drun_release() {
```
ielashi, line 19: nit:
```suggestion
maybe_download_drun() {
```

ielashi (approved):
LGTM!

PR #258, opened 2023-10-06
fix: deserialize `BlockTree` iteratively

ielashi:
# Problem
We're now running into a situation where the Bitcoin testnet has ~125k unstable blocks due to a difficulty reset.
Recursively deserializing these blocks, which we do in `post_upgrade`, causes a stack overflow.

# Solution
Rewrite the `BlockTree` deserialization to be iterative rather than recursive. Running a test with the current production state confirms that it fixes that stack overflow issue.

### DIFF
--- a/canister/src/blocktree/serde.rs
+++ b/canister/src/blocktree/serde.rs
@@ -52,37 +52,41 @@ impl<'de> Visitor<'de> for BlockTreeDeserializer {
     }
 
     fn visit_seq<A: SeqAccess<'de>>(self, mut seq: A) -> Result<Self::Value, A::Error> {
-        fn next<'de, A: SeqAccess<'de>>(seq: &mut A) -> (Block, usize) {
+        fn next<'de, A: SeqAccess<'de>>(seq: &mut A) -> Option<(Block, usize)> {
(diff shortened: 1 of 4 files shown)


ielashi:
Marking this as a draft while I explore a safe alternative in Rust.

ielashi:
I updated the PR with an alternative implementation that doesn't use `unsafe`. cc @dsarlis

maksymar (approved):

canister/src/blocktree/serde.rs
```diff
@@ -52,21 +52,41 @@ impl<'de> Visitor<'de> for BlockTreeDeserializer {
...
+                stack.push((tree, children_to_add - 1));
+
+                // Add the child to the stack.
+                let (child, grand_children_to_add) = next(&mut seq).unwrap();
+                stack.push((BlockTree::new(child), grand_children_to_add));
+            }
         }
 
-        Ok(build_tree(&mut seq))
+        unreachable!();
```
dsarlis, line 90: Maybe also add a message so it's a bit more clear what case we hit?

dsarlis (approved):

PR #261, opened 2023-10-10
fix: bound length of chain on testnet

ielashi:
We've run into a state in production where there exists:

1. A very deep chain.
2. Small forks from that deep chain.

In a previous incident we've added safety measures to cap the length of the chain on testnet when it gets very deep due to a difficulty reset, but that safety measure only applied if there were no forks on that chain.

This PR generalizes the safety measure, which now kicks in even if the deep chain has small forks.

### DIFF
--- a/canister/src/unstable_blocks.rs
+++ b/canister/src/unstable_blocks.rs
@@ -1000,7 +1000,7 @@ mod test {
         );
 
         // If there's a very long testnet chain `A`, and there exists another chain `B` s.t.
-        // depth(A) - depth(B) <= TESTNET_CHAIN_MAX_DEPTH, the root of chain `A` is considered stable.
+        // depth(A) - depth(B) < TESTNET_CHAIN_MAX_DEPTH, the root of chain `A` is considered stable.
         assert_eq!(peek(&unstable_blocks), Some(&chain[0]));
 
         // Add one more block to the second chain, so that it's depth is `TESTNET_CHAIN_MAX_DEPTH`.
@@ -1011,7 +1011,7 @@ mod test {
         )
         .unwrap();
 
-        // Now, depth(A) - depth(B) > TESTNET_CHAIN_MAX_DEPTH and the root of chain `A`
+        // Now, depth(A) - depth(B) >= TESTNET_CHAIN_MAX_DEPTH and the root of chain `A`
         // is considered stable.
         assert_eq!(peek(&unstable_blocks), None);
     }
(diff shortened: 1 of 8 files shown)


canister/src/unstable_blocks.rs
```diff
@@ -14,7 +14,9 @@ use serde::{Deserialize, Serialize};
 mod next_block_headers;
 use self::next_block_headers::NextBlockHeaders;
 
-const TESTNET_MAX_SOLO_CHAIN_LENGTH: u128 = 1000;
+// The maximum number of blocks that a chain on testnet can exceed other chains before it's
```
THLO, line 17:
```suggestion
// The maximum number of blocks that a chain on testnet can exceed other chains before its
```

canister/src/unstable_blocks.rs
```diff
@@ -327,24 +329,38 @@ fn get_stable_child(blocks: &UnstableBlocks) -> Option<usize> {
...
+                        let second_deepest_depth = match depths.len().checked_sub(2) {
+                            None => 0,
+                            Some(idx) => {
+                                let (_, second_child_idx) = depths[idx];
+                                blocktree::depth(&blocks.tree.children[second_child_idx])
+                            }
+                        };
+
+                        // NOTE: We use a `saturating_sub` here because `depths` is ordered by
+                        // `difficult_based_depth`, whereas here the chains are compared by their
```
THLO, line 356:
```suggestion
                        // `difficulty_based_depth`, whereas here the chains are compared by their
```

canister/src/unstable_blocks.rs
```diff
@@ -937,4 +953,66 @@ mod test {
...
+
+        // The normalized stability threshold is still not met, which means that, in theory,
+        // there are no stable blocks that can be popped.
+        assert!(
+            unstable_blocks.blocks_difficulty_based_depth()
+                < unstable_blocks.normalized_stability_threshold()
+        );
+
+        // If there's a very long testnet chain `A`, and there exists another chain `B` s.t.
+        // depth(A) - depth(B) < TESTNET_CHAIN_MAX_DEPTH, the root of chain `A` is considered stable.
```
dsarlis, line 1003:
```suggestion
        // depth(A) - depth(B) < TESTNET_CHAIN_MAX_DEPTH, the root of chain `A` is considered unstable.
```

dsarlis (approved):

PR #263, opened 2023-10-16
feat: improve benchmarking reports

ielashi:
Replaces `criterion` with our own custom logic for reporting benchmarks. Our own benchmarking reporting offers a number of advantages over `criterion`:

* A simple file for persisting benchmark results across commits.
* The ability to track multiple measurements (e.g. instruction counts and memory)
* An option to persist the results at your leisure (criterion always persists)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -27,12 +27,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "anes"
-version = "0.1.6"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4b46cbb362ab8752921c97e041f5e366ee6297bd428a31275b9fcf1e380f7299"
-
-[[package]]
 name = "anstream"
 version = "0.3.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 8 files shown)


dragoljub-duric:
It looks LGTM to me. Just one question, you mentioned that when you change this, in the benchmark CI test we will not need to clone the current master to get the old results, in other words, we will always have the results of the benchmark of the current master attached to it. But I don't know where we are saving results, can you please explain this to me? I know we can run this locally and then attach results to the commit, but that may introduce the problem that the results attached to the master are not updated.

ielashi:
> Just one question, you mentioned that when you change this, in the benchmark CI test we will not need to clone the current master to get the old results, in other words, we will always have the results of the benchmark of the current master attached to it. But I don't know where we are saving results, can you please explain this to me? I know we can run this locally and then attach results to the commit, but that may introduce the problem that the results attached to the master are not updated.

@dragoljub-duric With this PR, we're now storing the benchmark results in the results.yml file [here](https://github.com/dfinity/bitcoin-canister/pull/263/files#diff-2ce8f86258f915e0aa3e3db846e514c99937fa2f443b83d29d2670329d7ec32cR1). The goal with the CI changes is to ensure that the `results.yml` has been updated with the latest performance reports (which the developer can do using the command `cargo bench -- --persist`). To detect if the performance data is out of date, we can do the following:

1. Run `cargo bench` on the PR. This will output a report comparing the performance of the PR compared to the results stored in `results.yml`.
2. The report outputs whether or not a regression or an improvement has been detected. In which case, we are sure that the developer did not update `results.yml`, and the CI job can fail.

Does that make sense?

PR #268, opened 2023-11-16
feat: Use the burn cycles API in the Bitcoin canister

dragoljub-duric:


canister/src/heartbeat.rs
```diff
@@ -228,6 +232,30 @@ fn maybe_get_successors_request() -> Option<GetSuccessorsRequest> {
...
+#[cfg(not(target_arch = "wasm32"))]
+fn cycles_burn() {
+    let cycles_burnt = 1_000_000;
+    with_state_mut(|s| {
+        if let Some(metric_cycles_burnt) = &mut s.metrics.cycles_burnt {
+            *metric_cycles_burnt += cycles_burnt;
+        } else {
+            s.metrics.cycles_burnt = Some(cycles_burnt);
+        }
+    });
```
ielashi, line 256: Do we need this logic at all? I don't see anywhere where this is being triggered/used.

canister/src/heartbeat.rs
```diff
@@ -228,6 +232,30 @@ fn maybe_get_successors_request() -> Option<GetSuccessorsRequest> {
     })
 }
 
+#[cfg(target_arch = "wasm32")]
```
ielashi, line 235: We encapsulate all the system api calls in `runtime.rs`, so I'd move this logic there.

canister/src/heartbeat.rs
```diff
@@ -17,6 +17,10 @@ use ic_btc_types::{Block, BlockHash};
 /// The heartbeat fetches new blocks from the bitcoin network and inserts them into the state.
 pub async fn heartbeat() {
     print("Starting heartbeat...");
+
+    cycles_burn();
+    print("Done burning received cycles.");
```
ielashi, line 22: I'd remove this print statement, as it can quite noisy and burning cycles isn't very important for us from a logging perspective - we can already monitor it in metrics.

ielashi (commented):
The logic LGTM, but don't we need to add a test as well?

canister/src/heartbeat.rs
dragoljub-duric: I am not sure what you are referring to, but if you think about having optional field - we need an optional field because if it is nonoptional deserialization will fail because we have a new field.

canister/src/heartbeat.rs
dragoljub-duric: Done.

dragoljub-duric:
@maksymar can you please look at watchdog_helth_status test? The test is failing but I am not sure why.

dragoljub-duric:
@ielashi the test is added.

canister/src/runtime.rs
```diff
@@ -217,3 +218,24 @@ pub fn time() -> u64 {
...
+#[cfg(target_arch = "wasm32")]
+pub fn cycles_burn() {
+    let cycles_burnt = ic_cdk::api::cycles_burn(ic_cdk::api::canister_balance128());
+    add_cycles_burnt_to_metric(cycles_burnt);
+}
+
+#[cfg(not(target_arch = "wasm32"))]
+pub fn cycles_burn() {
+    add_cycles_burnt_to_metric(1_000_000);
+}
```
ielashi, line 241: `runtime.rs` is basically a wrapper around calls to the cdk. It exists so that we can provide test-friendly alternatives to SDK calls, and it shouldn't include any other logic. I'd do this:

1. Constrain this logic to only the system API call:
```
#[cfg(target_arch = "wasm32")]
pub fn cycles_burn() -> u128 {
    ic_cdk::api::cycles_burn(ic_cdk::api::canister_balance128())
}

#[cfg(not(target_arch = "wasm32"))]
pub fn cycles_burn() {
   1_000_000
}
```

The rest of the logic, like updating a metric, doesn't belong here. I'd keep that in `heartbeat.rs` as you had before.

canister/src/tests.rs
```diff
@@ -716,3 +716,35 @@ async fn test_syncing_with_next_block_headers() {
 
     assert!(catch_unwind(verify_synced).is_err());
 }
+
+#[async_std::test]
+async fn test_cycles_burnt() {
```
ielashi, line 721: This test is testing that we are incrementing the metrics, but it's not actually testing that we're burning cycles. I think we'd need an e2e test for that, otherwise we'll never exercise the code path where real cycles are actually burnt.

canister/src/runtime.rs
dragoljub-duric: Done.

canister/src/tests.rs
dragoljub-duric: The test is added now, this was the best I could do it in order not to make the test flaky. It confirms that cycles are burnt, after initializing canister with a lot of cycles.

dragoljub-duric: PTAL

e2e-tests/cycles_burn.sh
```diff
@@ -0,0 +1,45 @@
...
+    send_transaction_base = 0;
+    send_transaction_per_byte = 0;
+  };
+  api_access = variant { enabled };
+  disable_api_if_not_fully_synced = variant { enabled };
+  watchdog_canister = null;
+})"
+
+# Check that cycles are burnt.
+if [ "$(get_balance)" != "0" ]; then
```
ielashi, line 41: Maybe it'd be more robust to sleep here for a few seconds before calling get balance. Otherwise, it's possible to execute this call before the first heartbeat of the canister runs.

canister/src/tests.rs
ielashi, line 721:
```suggestion
async fn cycles_burnt_are_tracked_in_metrics() {
```

ielashi (approved):

canister/src/heartbeat.rs
```diff
@@ -17,6 +17,10 @@ use ic_btc_types::{Block, BlockHash};
 /// The heartbeat fetches new blocks from the bitcoin network and inserts them into the state.
 pub async fn heartbeat() {
     print("Starting heartbeat...");
+
+    let cycles_burnt = cycles_burn();
```
ielashi, line 21:
```suggestion
    // Burn any cycles in the canister's balance (to count towards the IC's cycles burn rate)
    let cycles_burnt = cycles_burn();
```

PR #272, opened 2023-11-29
chore: Update candid and ic-cdk

frederikrothenberger:
This PR upgrades `candid` and the `ic-cdk` (which also have a dependency on `candid`) to the latest version.
Updating the dependencies in this repo is a blocker to update them in the ic repo.


ielashi (approved):
LGTM

PR #270, opened 2023-11-22
chore: reduce min_explorers to 2 in tests setup

maksymar:
This PR fixes `watchdog_health_status` test by dropping `min_explorers` requirement to 2 in test config.

The reason for the failure is that there's no IPv4 support in dfx for system subnets, which causes only 2 explorers (supposedly with IPv6) to have http data which was not enough for the test to pass (old requirement was to have at least 3 explorers). Dropping the requirement temporarily fixes the test.

This PR is a short-term workaround. Long-term it'd require either a support for IPv4 or more explorers with IPv6.

Response example where `api_blockcypher_com_mainnet` and `chain_api_btc_com_mainnet` have responses:
```
  record {
    height_target = null;
    height_status = variant { not_enough_data };
    explorers = vec {
      record {
        height = null;
        provider = variant { api_blockchair_com_mainnet };
      };
      record {
        height = opt (818_978 : nat64);
        provider = variant { api_blockcypher_com_mainnet };
      };
      record { height = null; provider = variant { blockchain_info_mainnet } };
      record { height = null; provider = variant { blockstream_info_mainnet } };
      record {
        height = opt (818_978 : nat64);
        provider = variant { chain_api_btc_com_mainnet };
      };
    };
    height_source = opt (818_978 : nat64);
    height_diff = null;
  },
```

Links:
- PASS https://ready.chair6.net/?url=https%3A%2F%2Fapi.blockcypher.com%2Fv1%2Fbtc%2Fmain
- PASS https://ready.chair6.net/?url=https%3A%2F%2Fchain.api.btc.com%2Fv3%2Fblock%2Flatest
- FAIL https://ready.chair6.net/?url=https%3A%2F%2Fapi.blockchair.com%2Fbitcoin%2Fstats

PR #277, opened 2024-01-19
fix: make the bitcoin testnet watchdog less sensitive

ielashi:
Sets the threshold for blocks ahead/behind from +/- 2 to +/- 100 for bitcoin testnet given that bitcoin testnet often has surges in blocks when the difficulty is low.

The configs of the watchdog canister were not being maintained in this repo, and instead were only maintained in the NNS proposals. After a discussion with @maksymar, we agreed to maintain the configurations here instead to make them easier to review and track.

As a result, the watchdog canister has now been updated to require a bitcoin network when being initialized, and based on that network it'll use the hard-coded bitcoin mainnet/testnet config.

### DIFF
--- a/watchdog/e2e-tests/utils.sh
+++ b/watchdog/e2e-tests/utils.sh
@@ -2,8 +2,6 @@
 
 # Function to deploy the watchdog canister for mainnet bitcoin_canister.
 deploy_watchdog_canister_mainnet() {
-  BITCOIN_NETWORK=mainnet
-  BITCOIN_CANISTER_ID=ghsi2-tqaaa-aaaan-aaaca-cai
   dfx deploy --no-wallet watchdog --argument "(variant { mainnet })"
 }
 
--- a/watchdog/e2e-tests/utils.sh
+++ b/watchdog/e2e-tests/utils.sh
@@ -4,22 +4,7 @@
 deploy_watchdog_canister_mainnet() {
   BITCOIN_NETWORK=mainnet
   BITCOIN_CANISTER_ID=ghsi2-tqaaa-aaaan-aaaca-cai
-  dfx deploy --no-wallet watchdog --argument "(opt record {
-    bitcoin_network = variant { ${BITCOIN_NETWORK} };
-    blocks_behind_threshold = 2;
-    blocks_ahead_threshold = 2;
-    min_explorers = 2;
-    bitcoin_canister_principal = principal \"${BITCOIN_CANISTER_ID}\";
-    delay_before_first_fetch_sec = 1;
-    interval_between_fetches_sec = 300;
-    explorers = vec {
-      variant { api_blockchair_com_mainnet };
-      variant { api_blockcypher_com_mainnet };
-      variant { blockchain_info_mainnet };
-      variant { blockstream_info_mainnet };
-      variant { chain_api_btc_com_mainnet };
-    };
-  })"
+  dfx deploy --no-wallet watchdog --argument "(variant { mainnet })"
 }
 
 # Function to get watchdog canister metrics.
--- a/watchdog/e2e-tests/utils.sh
+++ b/watchdog/e2e-tests/utils.sh
@@ -43,7 +43,7 @@ check_health_status_fields() {
     "explorers"
   )
   
-  health_status=$(dfx canister call watchdog health_status --query)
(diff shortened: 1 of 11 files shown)


watchdog/e2e-tests/utils.sh
```diff
@@ -2,24 +2,7 @@
 
 # Function to deploy the watchdog canister for mainnet bitcoin_canister.
 deploy_watchdog_canister_mainnet() {
-  BITCOIN_NETWORK=mainnet
-  BITCOIN_CANISTER_ID=ghsi2-tqaaa-aaaan-aaaca-cai
-  dfx deploy --no-wallet watchdog --argument "(opt record {
-    bitcoin_network = variant { ${BITCOIN_NETWORK} };
-    blocks_behind_threshold = 2;
-    blocks_ahead_threshold = 2;
-    min_explorers = 2;
```
maksymar, line 11: This can still be a problem for e2e-tests, since dfx of newer versions do not support IPv4.
See context in https://github.com/dfinity/bitcoin-canister/pull/270

watchdog/src/config.rs
```diff
@@ -101,8 +83,8 @@ impl Config {
     pub fn testnet() -> Self {
         Self {
             bitcoin_network: BitcoinNetwork::Testnet,
-            blocks_behind_threshold: BLOCKS_BEHIND_THRESHOLD,
-            blocks_ahead_threshold: BLOCKS_AHEAD_THRESHOLD,
+            blocks_behind_threshold: 100,
+            blocks_ahead_threshold: 100,
             min_explorers: MIN_EXPLORERS,
             bitcoin_canister_principal: Principal::from_text(TESTNET_BITCOIN_CANISTER_PRINCIPAL)
                 .unwrap(),
```
maksymar, line 90: We should remove `BitcoinBlockApi::ApiBitapsComTestnet` from testnet explorers since it was misbehaving for 30+ days and stopped providing data recently at all.

image

watchdog/src/config.rs
```diff
@@ -67,20 +57,12 @@ pub struct Config {
...
     pub fn mainnet() -> Self {
         Self {
             bitcoin_network: BitcoinNetwork::Mainnet,
-            blocks_behind_threshold: BLOCKS_BEHIND_THRESHOLD,
-            blocks_ahead_threshold: BLOCKS_AHEAD_THRESHOLD,
+            blocks_behind_threshold: 2,
+            blocks_ahead_threshold: 2,
             min_explorers: MIN_EXPLORERS,
             bitcoin_canister_principal: Principal::from_text(MAINNET_BITCOIN_CANISTER_PRINCIPAL)
                 .unwrap(),
```
maksymar, line 68: FYI: 
in the last upgrade config did not have `ApiBitapsComMainnet` endpoint because it was misbehaving and was removed (probably before that).
Just [checked](https://api.bitaps.com/btc/v1/blockchain/block/last) it now and it seems to be working again, so we can add it again to the default config.

https://dashboard.internetcomputer.org/proposal/125325

watchdog/e2e-tests/utils.sh
ielashi: I addressed with rust features, so that we don't reduce it in production as well.

watchdog/src/config.rs
ielashi: We have that already in this config, so no action required, right?

watchdog/src/config.rs
ielashi: Sure, maybe then we reduce the min explorers for testnet from 3 to 2?

watchdog/src/config.rs
maksymar: Yes, no action required for re-enabling that explorer that looks ok at the moment.

watchdog/src/config.rs
maksymar: SGTM reducing to 2 for the testnet.

watchdog/e2e-tests/utils.sh
maksymar: nice! thank you.

watchdog/src/config.rs
ielashi: Done.

maksymar (approved):

PR #282, opened 2024-02-09
feat: use canbench for benchmarks

ielashi:


maksymar (approved):

scripts/canbench_ci_run_benchmark.sh
```diff
@@ -0,0 +1,79 @@
...
+
+# Verify that canbench results are available.
+if [ ! -f "$CANBENCH_RESULTS_FILE" ]; then
+    echo "$CANBENCH_RESULTS_FILE not found. Did you forget to run \`canbench --persist\`?";
+    exit 1
+fi
+
+# Detect if canbench results file is up to date.
+pushd "$CANISTER_PATH"
+canbench --less-verbose > $CANBENCH_OUTPUT
```
dragoljub-duric, line 32: If I understand correctly, this one should run a benchmark against the master branch. Am I right? If so, please tell me when we checkout to the master branch?

maksymar: Actually I also got confused by the sequence. Seems the first call of `canbench` should be preceded with `pushd "$MAIN_BRANCH_DIR"`.

ielashi: We run canbench twice. The first time it's to check if `canbench_results.yml` in the current branch is up to date, that's this call. The second time happens on line 57 below, and that's where we compare with the master branch.

.github/workflows/workflow.yml
```diff
@@ -386,11 +386,20 @@ jobs:
             bash e2e-tests/cycles_burn.sh
 
   benchmark:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
     needs: cargo-build
-
+    env:
+      PROJECT_DIR: ./
```
maksymar, line 392: should it be just `.`? otherwise it gets propagated to all the variables downstream and if they have leading slash it translates into double slashes. it doesn't do harm, but it looks odd

```sh
+ CANISTER_PATH=./
+ COMMENT_MESSAGE_PATH=/tmp/canbench_comment_message.txt
+ MAIN_BRANCH_DIR=_canbench_main_branch
+ CANBENCH_OUTPUT=/tmp/canbench_output.txt
+ CANBENCH_RESULTS_FILE=.//canbench_results.yml
+ MAIN_BRANCH_RESULTS_FILE=_canbench_main_branch/.//canbench_results.yml
```

ielashi: Done.

scripts/canbench_ci_run_benchmark.sh
maksymar: ah, I see. I was misled by all the `pushd`/`pop` folder switching (which is a no-op in this case, since the canister dir is the same is root) -- I thought `canbench` is run in two branches.

Now I see that the first run is to check that the current branch has committed an up-to-date `canbench_results.yml` and then copy results from the main branch and run it again for comparison.

PR #286, opened 2024-04-03
feat (EXC-1590): make burning cycles configurable

ielashi:
# Problem
The latest version of the Bitcoin canister automatically burns all cycles it receives. DFX is unable to upgrade to that version since deploying it on a local app subnet immediately results in it being uninstalled.

# Solution
Make burning cycles configurable, such that DFX can disable it when deploying the Bitcoin canister locally.

### DIFF
--- a/canister/candid.did
+++ b/canister/candid.did
@@ -35,6 +35,7 @@ type config = record {
   api_access : flag;
   disable_api_if_not_fully_synced : flag;
   watchdog_canister : opt principal;
+  burn_cycles : flag;
 };
 
 type fees = record {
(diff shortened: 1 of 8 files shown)


canister/candid.did
```diff
@@ -35,6 +35,7 @@ type config = record {
   api_access : flag;
   disable_api_if_not_fully_synced : flag;
   watchdog_canister : opt principal;
+  burn_cycles : flag;
```
dsarlis, line 38: Is this safe? If we don't make it optional would a canister upgrade work fine? I suppose it would because this flag is only used in a request when setting the config (below that request has the flag as an optional field) and the only other place we persist it is in the canister's state which should work given the `#[serde(default)]` has been added?

ielashi: Yes, exactly. We've done this before with other configs.

dsarlis (approved):

PR #294, opened 2024-04-26
chore: reduce block ingestion instruction limit from 2B to 1B

ielashi:
Reducing the amount of instructions done per round when ingesting blocks to 1B should result in a smoother finalization rate on the subnet.

### DIFF
--- a/canister/src/state.rs
+++ b/canister/src/state.rs
@@ -400,7 +400,7 @@ mod test {
 
         // Ingest the next block. This time, the performance counter is set so that
         // the ingestion is time-sliced.
-        crate::runtime::set_performance_counter_step(1_000_000_000);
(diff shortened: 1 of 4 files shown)


dsarlis (approved):

canister/src/state.rs
```diff
@@ -400,7 +400,7 @@ mod test {
 
         // Ingest the next block. This time, the performance counter is set so that
         // the ingestion is time-sliced.
-        crate::runtime::set_performance_counter_step(1_000_000_000);
+        crate::runtime::set_performance_counter_step(100_000_000);
```
dragoljub-duric, line 403: What is the reason we decreased the counter step here 10 times, why everywhere else it is halved?

PR #297, opened 2024-05-15
feat: Add stub for Bitcoin headers endpoint

dragoljub-duric:
As stated in [here](https://github.com/dfinity/interface-spec/pull/298), candid interface should look as:

```
type bitcoin_block_header = blob;

type get_block_headers_request = record {
	start_height : nat32;
	end_height: opt nat32;
};

type get_block_headers_result = record {
	tip_height: nat32;
	block_headers: vec bitcoin_block_header;
};

bitcoin_get_block_headers : (get_block_headers_request) -> (get_block_headers_response);
```
This PR will add no-op endpoint for returning the block headers in the [follow-up](https://github.com/dfinity/bitcoin-canister/pull/298) we will implement specified function.


interface/src/lib.rs
```diff
@@ -346,6 +347,72 @@ pub enum GetUtxosError {
...
+/// The response returned for a request for getting the block headers from a given height.
+#[derive(CandidType, Debug, Deserialize, PartialEq, Eq, Clone)]
+pub struct GetBlockHeadersResponse {
+    pub tip_height: Height,
+    pub block_headers: Vec<BlockHeader>,
+}
+
+/// Errors when processing a `get_block_headers` request.
+#[derive(CandidType, Debug, Deserialize, PartialEq, Eq, Clone)]
+pub enum GetBlockHeadersError {
```
ielashi, line 366: Strictly speaking, this type is not part of the interface as it we don't return it in any of our endpoints and it isn't defined in the `.did` file, right? If yes, it would make sense to move this inside the `canister` crate as it'd be considered an implementation detail.

canister/src/api/get_block_headers.rs
```diff
@@ -0,0 +1,75 @@
+use ic_btc_interface::{GetBlockHeadersError, GetBlockHeadersRequest, GetBlockHeadersResponse};
+
+// Various profiling stats for tracking the performance of `get_block_headers`.
+#[derive(Default, Debug)]
+struct Stats {
```
ielashi, line 5: Is this struct being used at all? If not, I'd remove it and introduce it only whenever it becomes used.

canister/src/main.rs
```diff
@@ -76,6 +76,14 @@ pub fn bitcoin_get_utxos_query(request: GetUtxosRequest) {
     };
 }
 
+#[update(manual_reply = true)]
+pub fn bitcoin_get_block_headers(request: GetBlockHeadersRequest) {
```
ielashi, line 80: I'm wondering if this is a good opportunity to start using [PocketIC](https://crates.io/crates/pocket-ic). Perhaps we can have a test where we query this endpoint and verify that right now it traps with a message saying that it's unimplemented, and later that test can evolve as you build the functionality.

canister/src/api/get_block_headers.rs
```diff
@@ -0,0 +1,75 @@
...
+}
+
+#[cfg(test)]
+mod test {
+    use ic_btc_interface::{Config, GetBlockHeadersError, GetBlockHeadersRequest, Network};
+
+    use crate::api::get_block_headers;
+
+    #[test]
+    fn get_block_headers_malformed_heights() {
```
ielashi, line 51: nit: While this is already a good time, and related to my other comment, consider writing this with PocketIC instead. This unit test is only testing this method, whereas PocketIC would test the end-to-end flow, which would make the test stronger.

ielashi (commented):
Looks good! Some minor nitpicks. Also, I'd avoid linking to private documentation in the PR description as external readers can't access them. You can instead link to the public documentation: https://github.com/dfinity/interface-spec/pull/298

canister/src/api/get_block_headers.rs
dragoljub-duric: Done.

interface/src/lib.rs
dragoljub-duric: That is a valid approach. But we currently have `GetUtxosError`, `SendTransactionError`, and `GetBalanceError` to which all of the things you mentioned apply and it is still in `interface/src/lib.rs`. I think we should keep it consistent, so how about we keep `GetBlockHeadersError ` as is in this PR, and if you still think we should move it to canister, in follow-up I can move all types from above to canister?

canister/src/main.rs
dragoljub-duric: How about I add tests with PocketIC in the follow-up?

canister/src/api/get_block_headers.rs
dragoljub-duric: Same as the comment above. In either case I think that unit test is fine to stay.

PR #298, opened 2024-05-17
feat: [EXC-1619] Implement `get_block_headers()` endpoint

dragoljub-duric:
This PR implements `get_block_headers` endpoint as stated in https://github.com/dfinity/interface-spec/pull/298.
E2E test will be added in follow-up.
Charging cycles, and determining the cost of the endpoint will be added in follow-up.


canister/src/api/get_block_headers.rs
```diff
@@ -1,17 +1,139 @@
...
+
+            vec_headers = with_state(|s| {
+                let block_heights = &s.stable_block_headers.block_heights;
+                let block_headers = &s.stable_block_headers.block_headers;
+                block_heights
+                    .range(start_height..=end_range_in_stable_blocks)
+                    .map(|(_, block_hash)| block_headers.get(&block_hash).unwrap().into())
+                    .collect()
+            });
+        }
```
ielashi, line 94: These checks feel more complicated than they need to be. Can't we just do this?

```suggestion
    let mut vec_headers =  with_state(|s| {
                let block_heights = &s.stable_block_headers.block_heights;
                let block_headers = &s.stable_block_headers.block_headers;
                block_heights
                    .range(start_height..=end_height)
                    .map(|(_, block_hash)| block_headers.get(&block_hash).unwrap().into())
                    .collect()
            });
```

canister/src/api/get_block_headers.rs
```diff
@@ -1,17 +1,139 @@
...
     }
-    Ok(())
+
+    // How the last stable block is located in unstable_blocks, there will always
+    // be the block in unstable_blocks.
+    let height_of_first_block_in_unstable_blocks =
+        height_of_last_block_in_stable_blocks.map_or(0, |h| h + 1);
```
ielashi, line 100: I think this can be made much clearer to read:
```suggestion
    let height_of_first_block_in_unstable_blocks = state.stable_height();
```

canister/candid.did
```diff
@@ -52,6 +52,9 @@ type fees = record {
   get_current_fee_percentiles_maximum : nat;
   send_transaction_base : nat;
   send_transaction_per_byte : nat;
+  get_block_headers_base : nat;
+  get_block_headers_cycles_per_ten_instructions : nat;
+  get_block_headers_maximum : nat;
```
ielashi, line 57: Here we're assuming that we will charge based on the number of instructions it takes to process the request. That's not necessarily how we want to charge. We may prefer for instance to charge based on the number of headers retrieved. I'd suggest removing anything related to fees from this PR and we can add it once we know how we want to charge.

How we want to charge can be a separate discussion with @THLO.

canister/src/api/get_block_headers.rs
```diff
@@ -1,17 +1,139 @@
...
+    // The total number of instructions used to process the request.
+    ins_total: u64,
+
+    // The number of instructions used to build the block headers vec.
+    ins_build_block_headers_vec: u64,
+}
+
+fn verify_requested_height_range_and_return_effective_range(
+    request: &GetBlockHeadersRequest,
+) -> Result<(u32, u32), GetBlockHeadersError> {
```
ielashi, line 23: What happens if a user requests 100k headers? We need to bound how many headers are returned, otherwise we'll run out of instructions. I suggest bounding it to 100 for now. @THLO WDYT?

canister/src/api/get_block_headers.rs
```diff
@@ -62,4 +233,313 @@ mod test {
...
+            prev_block_header = *block.header();
+
+            let mut block_blob = vec![];
+            block.header().consensus_encode(&mut block_blob).unwrap();
+            blobs.push(block_blob);
+
+            with_state_mut(|state| insert_block(state, block).unwrap());
+        }
+
+        with_state_mut(ingest_stable_blocks_into_utxoset);
```
dragoljub-duric, line 455: Proptest is failing on this with: 
`Cannot insert outpoint OutPoint { txid: 4a5e1e4baab89f3a32518a88c31bc87f618f76673e2cc77ab2127b7afdeda33b, vout: 0 }`
@ielashi did you see something similar, or do you know what can cause it?

canister/src/api/get_block_headers.rs
dragoljub-duric: Done.

canister/src/api/get_block_headers.rs
dragoljub-duric: Done.

canister/candid.did
dragoljub-duric: Done.

canister/src/api/get_block_headers.rs
dragoljub-duric: Done.

dragoljub-duric: PTAL.

canister/src/api/get_block_headers.rs
```diff
@@ -1,17 +1,128 @@
...
+        let block_heights = &s.stable_block_headers.block_heights;
+        let block_headers = &s.stable_block_headers.block_headers;
+        block_heights
+            .range(start_height..=end_height)
+            .map(|(_, block_hash)| block_headers.get(&block_hash).unwrap().into())
+            .collect()
+    });
+
+    // How the last stable block is located in `unstable_blocks`, the height of the
+    // first block in `unstable_blocks` is equal to `stable_height`.
```
ielashi, line 88:
```suggestion
    // The last stable block is located in `unstable_blocks`, the height of the
    // first block in `unstable_blocks` is equal to `stable_height`.
```

canister/src/api/get_block_headers.rs
```diff
@@ -1,17 +1,128 @@
...
+#[derive(Default, Debug)]
+struct Stats {
+    // The total number of instructions used to process the request.
+    ins_total: u64,
+
+    // The number of instructions used to build the block headers vec.
+    ins_build_block_headers_vec: u64,
+}
+
+fn verify_requested_height_range_and_return_effective_range(
```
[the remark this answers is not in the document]
dragoljub-duric: Done.

canister/src/api/get_block_headers.rs
```diff
@@ -1,17 +1,128 @@
...
+    request: &GetBlockHeadersRequest,
+) -> Result<(GetBlockHeadersResponse, Stats), GetBlockHeadersError> {
+    let (start_height, mut end_height) =
+        verify_requested_height_range_and_return_effective_range(request)?;
+
+    // Bound the length of block headers vec.
+    end_height = std::cmp::min(
+        end_height,
+        start_height + MAX_BLOCK_HEADERS_PER_RESPONSE - 1,
+    );
```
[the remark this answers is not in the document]
dragoljub-duric: Done.

canister/src/api/get_block_headers.rs
```diff
@@ -1,17 +1,128 @@
...
+const MAX_BLOCK_HEADERS_PER_RESPONSE: u32 = 100;
+
+// Various profiling stats for tracking the performance of `get_block_headers`.
+#[derive(Default, Debug)]
+struct Stats {
+    // The total number of instructions used to process the request.
+    ins_total: u64,
+
+    // The number of instructions used to build the block headers vec.
+    ins_build_block_headers_vec: u64,
```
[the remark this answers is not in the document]
dragoljub-duric: WDYT to change it to measure instructions for extracting block headers from unstable blocks?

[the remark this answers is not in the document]
ielashi: Sounds good! We can measure both extracting stable and unstable blocks as well for completeness?

canister/src/api/get_block_headers.rs
```diff
@@ -1,17 +1,130 @@
...
+
+    // Add requested block headers located in stable_blocks.
+    let mut vec_headers: Vec<Vec<u8>> = with_state(|s| {
+        let block_heights = &s.stable_block_headers.block_heights;
+        let block_headers = &s.stable_block_headers.block_headers;
+        block_heights
+            .range(start_height..=end_height)
+            .map(|(_, block_hash)| block_headers.get(&block_hash).unwrap().into())
+            .collect()
+    });
```
ielashi, line 87: I thought I wrote this yesterday. `block_heights` and `block_headers` in the `BlockHeaderStore` are supposed to be private. To that end, I think we can refactor this code to be a method in `BlockHeaderStore` instead. For example, it could have the following signature:

```
    pub fn get_block_headers_in_range(
        &self,
        heights: std::ops::RangeInclusive<Height>,
    ) -> impl Iterator<Item = BlockHeaderBlob> + '_ {
```

canister/src/api/get_block_headers.rs
```diff
@@ -1,17 +1,130 @@
...
+            for i in start_range_in_unstable_blocks..=end_range_in_unstable_blocks {
+                let mut header_blob = vec![];
+                unstable_blocks[i as usize]
+                    .header()
+                    .consensus_encode(&mut header_blob)
+                    .unwrap();
+                vec_headers.push(header_blob);
+            }
+        });
+    }
```
ielashi, line 116: Same here. Try to encapsulate as much logic in the respective components as possible. In this case, this logic can be added into the `UnstableBlocks` component. Perhaps something like this:

```
    pub fn get_block_headers_in_range(
        &self,
        stable_height: Height,
        heights: std::ops::RangeInclusive<Height>,
    ) -> impl Iterator<Item = &BlockHeader> {

```

canister/src/api/get_block_headers.rs
```diff
@@ -62,4 +210,356 @@ mod test {
...
+                    Some(start_height + length - 1)
+                } else {
+                    None
+                };
+                check_response(&blobs, start_height, end_height, block_num);
+            }
+        );
+    }
+
+    #[ignore]
```
ielashi, line 550: Is it intentional to leave this test ignored?

canister/src/api/get_block_headers.rs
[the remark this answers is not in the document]
dragoljub-duric: Good idea, done.

canister/src/api/get_block_headers.rs
dragoljub-duric: Done.

canister/src/api/get_block_headers.rs
dragoljub-duric: Done.

canister/src/block_header_store.rs
```diff
@@ -71,6 +71,14 @@ impl BlockHeaderStore {
                 .expect("block header must exist")
         })
     }
+
+    /// Returns block headers in the range [start_hegiht, end_height].
```
ielashi, line 75: typo

canister/src/block_header_store.rs
```diff
@@ -71,6 +71,14 @@ impl BlockHeaderStore {
...
+    pub fn get_block_headers_in_range(&self, start_height: u32, end_height: u32) -> Vec<Vec<u8>> {
```
ielashi, line 76: All the methods in this component are dealing with block headers as either `BlockHeaderBlob` or `BlockHeader`. It's confusing/inconsistent to have this method now return a block header as `Vec<u8>`. That's why my recommendation is to have the following signature:

```
    pub fn get_block_headers_in_range(
        &self,
        heights: std::ops::RangeInclusive<Height>,
    ) -> impl Iterator<Item = BlockHeaderBlob> + '_ {
```

This function returns an iterator of `BlockHeaderBlob`, when then in `get_block_headers` you can `map` into a `Vec<u8>`.

canister/src/block_header_store.rs
```diff
@@ -85,3 +93,47 @@ fn init_block_headers() -> StableBTreeMap<BlockHash, BlockHeaderBlob, Memory> {
...
+#[cfg(test)]
+mod test {
+    use bitcoin::consensus::Encodable;
+    use proptest::proptest;
+
+    use crate::{block_header_store::BlockHeaderStore, test_utils::BlockBuilder};
+
+    #[test]
+    fn test_get_block_headers_in_range() {
+        let mut vec_headers = vec![];
```
ielashi, line 106: nit: `vec_` is redundant. It's clear that it's a vec

```suggestion
        let mut headers = vec![];
```

canister/src/unstable_blocks.rs
```diff
@@ -167,6 +167,24 @@ impl UnstableBlocks {
...
         chain
     }
+
+    /// Returns block headers in the inclusive range provided as an argument
+    /// relative to the start of unstable blocks.
+    pub(crate) fn get_block_headers_in_range(
+        &self,
+        start_range_in_unstable_blocks: u32,
+        end_range_in_unstable_blocks: u32,
+    ) -> Vec<Vec<u8>> {
```
ielashi, line 177: Two comments here:

1. Like all other methods in this component, we should return `BlockHeader` and not necessarily assume that the caller wants it to be encoded. This function maybe used for other purposes in the future.
2. I'd find it more intuitive if we let this method take the actual heights and not the unstable heights. The unstable heights can be computed within this function by passing in the `stable_height`. We already do this for other methods in this component.

The method signature I suggested addresses these two issues.

```
   pub fn get_block_headers_in_range(
        &self,
        stable_height: Height,
        heights: std::ops::RangeInclusive<Height>,
    ) -> impl Iterator<Item = &BlockHeader> {
```

dragoljub-duric: Done.

canister/src/block_header_store.rs
dragoljub-duric: Done.

canister/proptest-regressions/blocktree.txt
```diff
@@ -0,0 +1,7 @@
...
...9dd76d9df4a91913067f13fd47dca48420335cd64ed5ca6262bfe3ecbeb, time: 0, bits: 545259519, nonce: 2 }, txdata: [Transaction { version: 1, lock_time: 0, input: [TxIn { previous_output: OutPoint { txid: 0000000000000000000000000000000000000000000000000000000000000000, vout: 4294967295 }, script_sig: Script(), sequence: 4294967295, witness: Witness { content: [], witness_elements: 0, last: 0, second_to_last: 0 } }], output: [TxOut { value: 5000000000, script_pubkey: Script(OP_DUP OP_HASH160 OP_PUSHBYTES_20 1c06d2ab6d4557646aeda8477edc4143bb728b7f OP_EQUALVERIFY OP_CHECKSIG) }] }] }, transactions: [Transaction { tx: Transaction { version: 1, lock_time: 0, input: [TxIn { previous_output: OutPoint { txid: 0000000000000000000000000000000000000000000000000000000000000000, vout: 4294967295 }, script_sig: Script(), sequence: 4294967295, witness: Witness { content: [], witness_elements: 0, last: 0, second_to_last: 0 } }], output: [TxOut { value: 5000000000, script_pubkey: Script(OP_DUP OP_HASH160 OP_PUSHBYTES_20 1c06d2ab6d4557646aeda8477edc4143bb728b7f OP_EQUALVERIFY OP_CHECKSIG) }] }, txid: RefCell { value: None } }], block_hash: RefCell { value: None }, mock_difficulty: None }, children: [] } }
```
dragoljub-duric, line 7: @ielashi These two protests are failing locally but passing on CI, this started happening today, maybe after I merged the newest master to this branch.

canister/src/state.rs
```diff
@@ -113,6 +114,10 @@ impl State {
     pub fn get_utxos(&self, address: Address) -> AddressUtxoSet<'_> {
         AddressUtxoSet::new(address, &self.utxos, &self.unstable_blocks)
     }
+
+    pub fn get_unstable_blocks_in_main_chain(&self) -> BlockChain {
```
ielashi, line 118: I think this isn't used anywhere, right? If not, I'd remove it.

canister/src/api/get_block_headers.rs
```diff
@@ -1,17 +1,125 @@
...
+                return Err(GetBlockHeadersError::EndHeightDoesNotExist {
+                    requested: end_height,
+                    chain_height,
+                });
+            }
+            // If `end_height` is provided then it should be the
+            // end of effective height range.
+            (request.start_height, end_height)
+        } else {
+            // If `end_height`` is not provided then the end of effective
```
ielashi, line 58:
```suggestion
            // If `end_height` is not provided then the end of effective
```

canister/src/api/get_block_headers.rs
```diff
@@ -1,17 +1,125 @@
...
+            requested: request.start_height,
+            chain_height,
+        });
     }
-    Ok(())
+
+    let (effective_start_height, mut effective_end_height) =
+        if let Some(end_height) = request.end_height {
+            if end_height < request.start_height {
+                return Err(GetBlockHeadersError::StartHeightLagerThanEndHeight {
```
ielashi, line 42: typo:
```suggestion
                return Err(GetBlockHeadersError::StartHeightLargerThanEndHeight {
```

canister/proptest-regressions/blocktree.txt
ielashi: It's passing locally for me. Try running `cargo clean` and make sure you're on a clean branch.

e2e-tests/disable-api-if-not-fully-synced-flag.sh
```diff
@@ -104,6 +104,18 @@ if ! [[ $MSG = *"Canister state is not fully synced."* ]]; then
   exit 1
 fi
 
+# bitcoin_get_block_headers should panic.
```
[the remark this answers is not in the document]
dragoljub-duric: I believe it will not be a problem if it stays. AFAIK we agreed not to hide the method behind the feature flag, and just left the endpoint exposed, furthermore I will add the rest of the e2e test very soon, I believe there will be no releases prior to that, hence I think it will be time wasteful to remove this now, and to put it back later.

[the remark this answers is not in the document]
ielashi: My understanding was that, at least so far, we're not adding the endpoint until Manu confirms. We can wait until next week when he's back if you like.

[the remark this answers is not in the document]
dragoljub-duric: Can you please PTAL, I returned unimplemented whenever the endpoint is called. Alternatively, we can remove(or comment) it from canister/main.rs, but I believe this is enough.

ielashi (approved):
lgtm, thank you!

PR #300, opened 2024-05-28
test: add candid interface compatibility check

ielashi:
Adds checks to ensure that the candid interface and the canister's endpoints are compatible/in sync.


dsarlis (approved):
Lgtm. We can also consider adopting the new candid type selector that allows us to auto-generate Rust bindings from the did file.

dsarlis:
> Lgtm. We can also consider adopting the new candid type selector that allows us to auto-generate Rust bindings from the did file (and then the compatibility check is even obsolete).

There's some information [here](https://github.com/dfinity/candid/tree/master/tools/didc) on how you can generate Rust bindings (it's in beta testing).

ielashi:
Thanks, will take a look.

PR #303, opened 2024-05-29
feat: add config to eagerly evaluate fee percentiles

ielashi:
Fee percentiles are typically evaluated lazily, meaning that they are not computed unless they're requested. This can be problematic in local development where there are very few transactions. Consider the following scenario in local development:

1. User mines a block with a non-coinbase transaction
2. User mines a block with only a coinbase transaction

If developers now call the `get_current_fee_percentiles` endpoint, they'll get an empty vector. This is incorrect, as there has already been one transaction with fees that can be used to compute the fee percentiles. This behavior happens because locally all blocks are marked as stable immediately, and we only evaluate fee percentiles in unstable blocks.

To address this issue, this commit adds a new configuration `lazily_evaluate_fee_percentiles`, which allows us to configure the canister whether fee percentiles are lazily evaluated or not. For local development environment, they'd be evaluated eagerly, and lazily otherwise.

See [this forum post](https://forum.dfinity.org/t/get-current-fee-percentiles-returns-empty-array-after-mining-a-block-with-no-transactions) for an example of a user running into this error.


dsarlis (commented):
IIUC, the bitcoin canisters in production should have the new flag set to `Enabled` (instead of the default value of `Disabled`). I didn't see where this would be happening? I expected maybe to set it in the `post_upgrade` hook of the canister for example.

THLO:
> IIUC, the bitcoin canisters in production should have the new flag set to `Enabled` (instead of the default value of `Disabled`).

I agree.

canister/src/heartbeat.rs
```diff
@@ -200,6 +203,14 @@ fn maybe_process_response() {
     });
 }
 
+fn maybe_compute_fee_percentiles() {
+    if with_state(|s| s.lazily_evaluate_fee_percentiles == Flag::Enabled) {
+        return;
+    }
+
+    with_state_mut(|s| get_current_fee_percentiles_impl(s));
```
THLO, line 211: What's up with this clippy warning?

ielashi:
> IIUC, the bitcoin canisters in production should have the new flag set to `Enabled`

That is correct.

> (instead of the default value of `Disabled`). I didn't see where this would be happening? I expected maybe to set it in the `post_upgrade` hook of the canister for example. Or the idea is that we'll pass the configuration during the next upgrade?

The default value of flags is actually `Enabled`. See [here](https://github.com/dfinity/bitcoin-canister/blob/89c28bad3b63da38f3223ccc13b1a5acb9ebaf16/interface/src/lib.rs#L536), so there isn't anything further that needs to be done.

canister/src/heartbeat.rs
ielashi: Fixed.

dsarlis:
> The default value of flags is actually `Enabled`. See [here](https://github.com/dfinity/bitcoin-canister/blob/89c28bad3b63da38f3223ccc13b1a5acb9ebaf16/interface/src/lib.rs#L536), so there isn't anything further that needs to be done.

Ah thanks, somehow I missed that and assumed the default is `Disabled`.

dsarlis (approved):

PR #304, opened 2024-06-03
bench: measure memory usage when serializing unstable blocks

ielashi:
Adds a benchmark to measure how much heap memory is being used when upgrading with a large number of unstable blocks. This is the codepath that we'll need to optimize in order to avoid running out of heap memory in the future.

### DIFF
--- a/test-utils/src/lib.rs
+++ b/test-utils/src/lib.rs
@@ -1,8 +1,11 @@
+use bitcoin::blockdata::constants::genesis_block;
 use bitcoin::{
-    secp256k1::rand::rngs::OsRng, secp256k1::Secp256k1, util::uint::Uint256, Address, Block,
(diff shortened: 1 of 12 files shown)


test-utils/src/lib.rs
```diff
@@ -73,10 +76,49 @@ impl BlockBuilder {
             None => genesis(merkle_root),
         };
 
-        Block { header, txdata }
+        BitcoinBlock { header, txdata }
     }
 }
 
+/// Builds a random chain with the given number of block and transactions
+/// and starting with the Regtest genesis block.
```
dsarlis, line 84:
```suggestion
/// starting with the Regtest genesis block.
```

dsarlis (approved):

PR #309, opened 2024-06-07
chore: [EXC-1620] Charge cycles for using get_block_headers endpoint

dragoljub-duric:
This PR implements cycles charging for bitcoin_get_block_headers endpoint.

After discussion, we concluded that the best tradeoff is to use charging similar to the one in get_utxos with the fixed `base_cost` and variable part based on the number of instructions executed and bound their sum with a predetermined upper bound on the cost for one call of the endpoint.

When we expose the endpoint, the e2e test for charging cycles will be added in the [follow-up](https://github.com/dfinity/bitcoin-canister/pull/310).


ielashi (commented):
Unit tests please :) You can see how we unit test other endpoints for inspiration.

dragoljub-duric:
@ielashi Done, PTAL.

ielashi (approved):

PR #312, opened 2024-06-12
chore: [EXC-1649] Add network field to BitcoinGetBlockHeadersRequest

dragoljub-duric:


interface/src/lib.rs
```diff
@@ -352,6 +352,7 @@ pub enum GetUtxosError {
 pub struct GetBlockHeadersRequest {
     pub start_height: Height,
     pub end_height: Option<Height>,
+    pub network: NetworkInRequest,
```
dragoljub-duric, line 355: @ielashi  In order to preserve backward compatibility with version 0.2.1 we should probably make this field optional, but since nobody really used that version is there a better way (maybe to remove it). WDYT?

PR #313, opened 2024-06-21
[DO NOT MERGE BEFORE RELEASE] chore: update post_upgrade to work with new release

ielashi:
In #306 the pre_upgrade was updated to use less heap memory. This commit updates the post_upgrade to work with the latest version of the pre_upgrade.

This commit had to be merged separately, as it can only be merged once the canister is released.


ielashi:
We don't this PR anymore

PR #314, opened 2024-07-10
chore: Upgrade bitcoin to version 0.32.2

ninegua:
Upgrade bitcoin library to version 0.32.2. Notable changes are:
- A number of types now use new type instead of primitive types like u32.
- `U256` is no longer exported from bitcoin, and instead `primitive-types::U256` is used when it is needed.
- Because of changes to `Address` type, we have to use `assume_checked` when it is needed.
- Dependency on `getrandom` (due to `secp256k1` dependency) now requires `custom` flag to be used.
- Some other miscellaneous type and name changes.

### DIFF
--- a/canister/src/types.rs
+++ b/canister/src/types.rs
@@ -1,5 +1,5 @@
 use bitcoin::{
-    Address as BitcoinAddress, Network as BitcoinNetwork, Script, TxOut as BitcoinTxOut,
+    Address as BitcoinAddress, Amount, Network as BitcoinNetwork, ScriptBuf, TxOut as BitcoinTxOut,
 };
 use candid::CandidType;
 use ic_btc_interface::{
@@ -25,7 +25,7 @@ const EXPECTED_PAGE_LENGTH: usize = 72;
 /// A Bitcoin transaction's output.
 #[derive(Ord, PartialOrd, Eq, PartialEq, Clone, Debug, Serialize, Deserialize)]
 pub struct TxOut {
-    pub value: u64,
+    pub value: Amount,
     pub script_pubkey: Vec<u8>,
 }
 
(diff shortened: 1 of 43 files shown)


canister/src/types.rs
```diff
@@ -25,7 +25,7 @@ const EXPECTED_PAGE_LENGTH: usize = 72;
 /// A Bitcoin transaction's output.
 #[derive(Ord, PartialOrd, Eq, PartialEq, Clone, Debug, Serialize, Deserialize)]
 pub struct TxOut {
-    pub value: u64,
+    pub value: Amount,
```
ninegua, line 28: This seems to be a breaking change. Shall we revert to maintain compatibility?

ielashi:
As I mentioned on Slack, I'm not sure why we're doing this upgrade. AFAICT it's bringing risk with no particular gain.

ninegua:
> As I mentioned on Slack, I'm not sure why we're doing this upgrade. AFAICT it's bringing risk with no particular gain.

Yes, perfectly understood. I made the PR just in case it may become useful in the future should an upgrade is needed. In general, library maintainers could bring bug fixes, enhancement or new features, so we probably don't want to be left behind for too long. The rust library for bitcoin has a [changelog](https://github.com/rust-bitcoin/rust-bitcoin/blob/master/bitcoin/CHANGELOG.md) that may be useful.

ielashi:
> > As I mentioned on Slack, I'm not sure why we're doing this upgrade. AFAICT it's bringing risk with no particular gain.
> 
> Yes, perfectly understood. I made the PR just in case it may become useful in the future should an upgrade is needed. In general, library maintainers could bring bug fixes, enhancements or new features, so we probably don't want to be left behind for too long. The rust library for bitcoin has a [changelog](https://github.com/rust-bitcoin/rust-bitcoin/blob/master/bitcoin/CHANGELOG.md) that may be useful.

I see, thanks @ninegua! In that case, would it make sense to mark it as a draft, and the team can take a closer look whenever they have the bandwidth to test this upgrade? I don't the team has the resources to do this soon.

rumenov:
thanks @ninegua , the NET will definitely benefit from this upgrade

maksymar:
FYI: #349 upgraded bitcoin crate to 0.32.4.

PR #315, opened 2024-07-16
fix: canister bootstrap scripts

ielashi:
Enhancements and fixes to the bootstrapping process:

* Removes some unnecessary unwraps, as they are no longer required by newer versions of `ic-stable-structures`.
* Updates the `build-utxos` package to use a `FileMemory`, as now the state is too big to fit in RAM and results in OOM errors otherwise.
* Merges the `state-builder` and `main-state-builder` projects (not strictly required, but is a nice simplication)
* Updates uploader canister to take the chunk hashes via an environment variable. We cannot pass them has init arguments as they are bigger than the maximum payload size.
* Removed unused `main.rs` script in the `state-builder` crate
* Upgrades dfx to 0.21.0


dsarlis (commented):
There's some compilation error, I guess maybe you didn't push some recent change?

ielashi:
> There's some compilation error, I guess maybe you didn't push some recent change?

Should be fixed now.

dsarlis (approved):

PR #316, opened 2024-07-22
chore: enhancements for bootstrapping

ielashi:
1. `post_upgrade` hook now take an optional configuration. This allows the upgrade proposal after bootstrapping to set appropriate configs, particularly fees.
2. The `post_upgrade` needs to work with the bootstrapped canister and the current bitcoin mainnet canister. The bitcoin mainnet canister still has an old `pre_upgrade` hook. To address, a `legacy_preupgrade` feature has been added. When upgrading the bootstrapped testnet canister, this feature won't be enabled, but when upgrading the bitcoin mainnet canister, this feature will need to be enabled. This has the unfortunate consequence that in the upcoming release the bitcoin testnet and mainnet canister wasms won't match.

### DIFF
--- a/canister/src/blocktree/serde.rs
+++ b/canister/src/blocktree/serde.rs
@@ -64,10 +64,18 @@ impl<'de> Visitor<'de> for BlockTreeDeserializer {
 
     fn visit_seq<A: SeqAccess<'de>>(self, mut seq: A) -> Result<Self::Value, A::Error> {
         fn next<'de, A: SeqAccess<'de>>(seq: &mut A) -> Option<(Block, usize)> {
-            // TODO(EXC-1639): Update the logic here to deserialize a `BitcoinBlock`
-            // rather than a `Block` once the canister is released.
-            seq.next_element()
+            // A feature to upgrade from a release canister that has the old pre_upgrade
+            // that serialized the entire Block (as opposed to only the `BitcoinBlock` inside it).
+            #[cfg(feature = "legacy_preupgrade")]
+            return seq
+                .next_element()
+                .expect("reading next element must succeed");
+
+            #[cfg(not(feature = "legacy_preupgrade"))]
+            return seq
+                .next_element::<(BitcoinBlock, usize)>()
                 .expect("reading next element must succeed")
+                .map(|(b, n)| (Block::new(b), n));
         }
 
         // A stack containing a `BlockTree` along with how many children remain to be added to it.
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -331,6 +331,35 @@ jobs:
         run: |
           bash e2e-tests/set_config.sh
 
+  post_upgrade_config:
+    runs-on: ubuntu-latest
+    needs: cargo-build
+
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/cache@v4
+        with:
+          path: |
+            ~/.cargo/registry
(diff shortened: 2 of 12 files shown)


dsarlis:
This seems like it's adding some extra overhead both in code but also the bit where the binaries will be different on bitcoin testnet/mainnet canisters. I'm not sure I understand the benefit though. Can you explain why you think this is better than setting fees with e.g. a `SetConfig` proposal after upgrading to the bitcoin canister binary?

ielashi:
> This seems like it's adding some extra overhead both in code but also the bit where the binaries will be different on bitcoin testnet/mainnet canisters. I'm not sure I understand the benefit though. Can you explain why you think this is better than setting fees with e.g. a `SetConfig` proposal after upgrading to the bitcoin canister binary?

The binaries being different is orthogonal to the `SetConfig` changes. The reason the binaries will be different is because rebootstrapping will use the current `pre_upgrade`, while the bitcoin mainnet canister is still using the old `pre_upgrade`. I tried making the same code work for both upgrades, but it doesn't seem to be feasible with `serde` and `ciborium` unfortunately, so different binaries are inevitable in this case.

As for the `post_upgrade` config, the reason why I added it is because without it:
1. We'll need an additional proposal to call `SetConfig`.
2. Between the time where the canister is upgraded and the `SetConfig` proposal is executed, the config, primarily the fees will be incorrectly set to zero.

Adding the config in the post_upgrades addresses both of these shortcomings. I'm OK with just having a third proposal as well and dropping the config in `post_upgrade`, no strong opinion, but we'll still need the `legacy_preupgrade` feature flag.

canister/src/blocktree/serde.rs
```diff
@@ -64,10 +64,18 @@ impl<'de> Visitor<'de> for BlockTreeDeserializer {
...
+            #[cfg(feature = "legacy_preupgrade")]
+            return seq
+                .next_element()
+                .expect("reading next element must succeed");
+
+            #[cfg(not(feature = "legacy_preupgrade"))]
+            return seq
+                .next_element::<(BitcoinBlock, usize)>()
                 .expect("reading next element must succeed")
+                .map(|(b, n)| (Block::new(b), n));
```
dsarlis, line 78: Maybe it still doesn't work but what if we tried something like: attempt to deserialize a `BitcoinBlock` and if it succeeds we proceed, if it fails we fall back to the old style deserialization. Wouldn't that cover both the bitcoin testnet canister (the first deserialization would work) but also would work for the mainnet bitcoin canister?

dsarlis (approved):

canister/src/blocktree/serde.rs
ielashi: Absolutely. It's very easy in theory, but no matter what I tried, it doesn't work. Approaches I tried:

1. Try deserializing as a `BitcoinBlock`, and if that fails, try deserializing as a `Block`. The problem here is that calling `next_element` advances the iterator if it fails, and there's no way to move back in the iterator to be able to try deserializing the same element again.
2. Try deserializing as a `Vec<u8>`, then try deserializing to the different block types. That approach also didn't work.
3. Try deserializing as a `ciborium::Value`, then deserialize into the different block types. Also didn't work.

I'm surprised so easy in theory cannot be accomplished, but I didn't want to waste more time on this.

dsarlis: Ok, let's not spend more time then. Thanks for the extra details.

PR #317, opened 2024-07-25
fix: do not include canbench in production

ielashi:
## Problem
Running `pre_upgrade` in the bootstrapping scripts fails with an error that `stable64_size` is only available in canisters.

## Solution
The bootstrapping scripts do not run the canister code as Wasm, and do not have access to IC-level APIs. Upon further investigation, the reason `stable64_size` was needed was because we were accidentally including `canbench` into the compilation of the canister code - it should only be included when we're benchmarking.

To fix this, the `build-scripts.sh` has been updated to build using `--package` rather than `--bin` - that actually builds the canister from the correct root `Cargo.toml` file and therefore doesn't include any additional features that were required by other packages.

I used this opportunity to add a test using `PocketIC`.

### DIFF
--- a/scripts/build-canister.sh
+++ b/scripts/build-canister.sh
@@ -19,10 +19,10 @@ fi
 
 if [[ -z "$FEATURES" ]]; then
   # No features provided
-  cargo build -p $CANISTER --target "$TARGET" --release
+  cargo build -p "$CANISTER" --target "$TARGET" --release
 else
   # Features provided
-  cargo build -p $CANISTER --target "$TARGET" --release --features "$FEATURES"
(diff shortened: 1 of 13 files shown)


scripts/build-canister.sh
```diff
@@ -19,10 +19,10 @@ fi
 
 if [[ -z "$FEATURES" ]]; then
   # No features provided
-  cargo build --bin "$CANISTER" --target "$TARGET" --release
+  cargo build -p "$CANISTER" --target "$TARGET" --release
```
dsarlis, line 22: Maybe leave a comment here explaining why the `-p` option which I guess is non-standard when you want to build a binary.

dsarlis (approved):

PR #327, opened 2024-09-23
feat: improve height target calculation for watchdog canister

maksymar:
This PR makes calculation of height target for watchdog canister more strict.

- Before: The target height is the median of explorer heights if there are enough explorers (>= `min_explorers`).
- After: The target height is the median only if enough explorers (>= `min_explorers`) are within a defined range of the median.

Configs (currently in prod):
- `mainnet` -- explorers 6, min_explorers 3, blocks range [-2, 2]
- `testnet` -- explorers 3, min_explorers 2, blocks range [-1000, 1000]

Making the height target calculation more strict for the `testnet` is necessary because the `testnet` has fewer explorers and greater instability, leading to large block height divergences. A stricter calculation helps reduce noise and false positives by ensuring only explorers with closely aligned heights are considered valid, improving accuracy and stability.

Mainnet explorers: 
image
image

Testnet explorers: 
image
image

### DIFF
--- a/watchdog/src/health.rs
+++ b/watchdog/src/health.rs
@@ -56,6 +56,7 @@ pub fn health_status() -> HealthStatus {
     )
 }
 
+/// Returns the median if `min_explorers` are within the block range around it.
 fn calculate_height_target(
     heights: &[u64],
     min_explorers: usize,
@@ -66,10 +67,10 @@ fn calculate_height_target(
         return None;
     }
 
-    let median = median(heights)? as i64;
+    let threshold = median(heights)? as i64;
     let (lo, hi) = (
-        (median + blocks_behind_threshold).max(0) as u64,
-        (median + blocks_ahead_threshold).max(0) as u64,
+        threshold.saturating_add(blocks_behind_threshold) as u64,
+        threshold.saturating_add(blocks_ahead_threshold) as u64,
     );
     let valid_explorers = heights
         .iter()
@@ -77,8 +78,7 @@ fn calculate_height_target(
         .count();
 
     if valid_explorers >= min_explorers {
-        // Return the median if enough explorers are within range.
-        Some(median as u64)
+        Some(threshold as u64)
     } else {
         None
     }
--- a/watchdog/src/health.rs
+++ b/watchdog/src/health.rs
@@ -178,17 +178,15 @@ mod test {
 
     #[test]
     fn test_calculate_height_target_not_enough_explorers() {
-        // Within threshold.
         test_calculate_height_target(CalculateHeightTargetTestData {
-            heights: &[10, 12],
+            heights: &[10, 12], // Within threshold.
             min_explorers: 3,
             blocks_behind_threshold: -1,
             blocks_ahead_threshold: 1,
             expected: None,
         });
-        // Outside threshold.
         test_calculate_height_target(CalculateHeightTargetTestData {
-            heights: &[10, 13],
+            heights: &[10, 13], // Outside threshold.
             min_explorers: 3,
             blocks_behind_threshold: -1,
             blocks_ahead_threshold: 1,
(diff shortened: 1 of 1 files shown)


watchdog/src/health.rs
```diff
@@ -56,18 +56,44 @@ pub fn health_status() -> HealthStatus {
...
+    blocks_ahead_threshold: i64,
+) -> Option<u64> {
+    if heights.len() < min_explorers {
+        return None;
+    }
+
+    let threshold = median(heights)? as i64;
+    let (lo, hi) = (
+        threshold.saturating_add(blocks_behind_threshold) as u64,
+        threshold.saturating_add(blocks_ahead_threshold) as u64,
```
dragoljub-duric, line 73: I think it will be easier to understand if `blocks_behind_threshold` and `blocks_ahead_threshold` are `u64`, and then we have:
```
let (lo, hi) = (
        threshold.saturating_sub(blocks_behind_threshold) as u64,
        threshold.saturating_add(blocks_ahead_threshold) as u64,
);
```

It is harder to infer what `blocks_behind_threshold` when it is set to be `-1`.

watchdog/src/health.rs
```diff
@@ -117,14 +144,69 @@ mod test {
...
+    #[test]
+    fn test_calculate_height_target_not_enough_explorers() {
+        test_calculate_height_target(CalculateHeightTargetTestData {
+            heights: &[10],
+            min_explorers: 3,
+            blocks_behind_threshold: -1,
+            blocks_ahead_threshold: 1,
+            expected: None,
+        });
+    }
```
dragoljub-duric, line 188: NIT: Are `CalculateHeightTargetTestData` and `test_calculate_height_target()`necessary at all? I do not see that we make code shorter using them.

watchdog/src/health.rs
```diff
@@ -117,14 +144,69 @@ mod test {
...
+
+    #[test]
+    fn test_calculate_height_target_explorers_are_in_range() {
+        test_calculate_height_target(CalculateHeightTargetTestData {
+            heights: &[10, 11, 12],
+            min_explorers: 3,
+            blocks_behind_threshold: -1,
+            blocks_ahead_threshold: 1,
+            expected: Some(11),
+        });
```
dragoljub-duric, line 209: Maybe I will add a couple more of test variations, e.g.:

1. We have explorers only above the threshold that are in the range(lo, hi), and their number is < `min_explorers`
2. We have explorers only above the threshold that are in the range, and their number is >= `min_explorers`
3. We have explorers only below the threshold that are in range, and their number is < `min_explorers`
4. We have explorers only below the threshold that are in range, and their number is >= `min_explorers`

watchdog/src/health.rs
maksymar: The fact that `blocks_behind_threshold` is negative comes from comparing it to height diff between bitcoin canister and target height (it can be either positive or negative) and `get_blocks_behind_threshold()` has a corresponding comment.
Changing it back is too much of work which is out of the scope of this PR, so I'd like to leave it as is.

watchdog/src/health.rs
maksymar: It makes tests a bit more readable when you see the names of variables.

```rs
// Before.
#[test]
fn test_calculate_height_target_explorers_not_in_range() {
    assert_eq!(test_calculate_height_target(&[10, 20, 30], 3, -1, 1), None);
}

// After.
#[test]
fn test_calculate_height_target_explorers_not_in_range() {
    test_calculate_height_target(CalculateHeightTargetTestData {
        heights: &[10, 20, 30],
        min_explorers: 3,
        blocks_behind_threshold: -1,
        blocks_ahead_threshold: 1,
        expected: None,
    });
}
```

watchdog/src/health.rs
maksymar: Done.

dragoljub-duric (approved):

PR #335, opened 2024-10-08
feat: Add a flag to allow for uploading chunks from the end of the list

dsarlis:
This adds a flag to the uploader binary to allow for choosing whether the upload starts from the first chunk (default) or the last chunk. This can be handy to speed up the upload of state since once can start two instance in parallel, one uploading from the first chunk and the other from the last one.


maksymar (approved):

ielashi (approved):
We really should've done this years ago in hindsight :D Thanks for putting this together.

PR #341, opened 2024-11-25
chore: update bootstrap scripts

maksymar:


bootstrap/utils.sh
```diff
@@ -0,0 +1,44 @@
+#!/usr/bin/env bash
+#
+# Utility functions for Bitcoin scripts.
+set -euo pipefail
+
+# Validate the network input.
+validate_network() {
+    local network=$1
+    local valid_networks=("mainnet" "testnet")
+    if ! [[ " ${valid_networks[*]} " =~ $network ]]; then
```
berestovskyy, line 10: This could potentially match substrings like `net`. For more precise matching, consider using explicit comparison or a case statement: `case $network in mainnet | testnet) echo VALID ;; *) echo INVALID ;; esac`

berestovskyy (approved):

bootstrap/utils.sh
maksymar: Thank you for pointing this out. here's a fix https://github.com/dfinity/bitcoin-canister/pull/342.

PR #347, opened 2024-12-09
chore: bump up canbench to 0.1.8

maksymar:
This PR upgrades `canbench` crate to v.0.1.8 and `pocket-ic` to 6.0.0.

Canbench tests for this PR showed a regression for `pre_upgrade_with_many_unstable_blocks`, but that is expected since in `canbench` v.0.1.5 there was a fix for correctly accounting the instructions when writing stable memory ([release](https://github.com/dfinity/canbench/releases/tag/v0.1.5)), which is exactly what is happening in bitcoin canister pre_upgrade.

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -68,23 +68,23 @@ jobs:
           rustup default $RUST_VERSION
           rustup target add wasm32-unknown-unknown
 
-      # - name: Install PocketIC (mac)
-      #   if: runner.os == 'macOS'
-      #   run: |
-      #     wget https://github.com/dfinity/pocketic/releases/download/$POCKET_IC_SERVER_VERSION/pocket-ic-x86_64-darwin.gz
-      #     gzip -d pocket-ic-x86_64-darwin.gz
-      #     mv pocket-ic-x86_64-darwin pocket-ic
-      #     chmod +x pocket-ic
-      #     mv pocket-ic ./canister/
+      - name: Install PocketIC (mac)
+        if: runner.os == 'macOS'
+        run: |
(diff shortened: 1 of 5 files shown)


.github/workflows/workflow.yml
```diff
@@ -3,6 +3,7 @@ name: CI
 env:
   RUST_VERSION: 1.76.0
   DFX_VERSION: 0.21.0
+  POCKET_IC_VERSION: 7.0.0
```
maksymar, line 6: NOTE: for some reason having 6.0.0 was not enough and the test was failing with

```shell
Installing the bitcoin canister...
thread 'canbench_is_not_in_bitcoin_canister' panicked at /home/maksym/.cargo/registry/src/index.crates.io-6f17d22bba15001f/pocket-ic-6.0.0/src/lib.rs:1518:9:
Incompatible PocketIC server version: got pocket-ic-server 6.0.0; expected pocket-ic-server 7.0.0.
stack backtrace:```

so I used 7.0.0 here.

michael-weigelt: Yes, the server's version is not the library's version. At the moment they are off by one, which makes it a bit more confusing than if they were off by more.

michael-weigelt: So I think it would be good to call this constant POCKET_IC_SERVER_VERSION

michael-weigelt: Also, canbench already downloads the correct version for you. Does this repo have its own suite of pocketic tests?

maksymar: > I think it would be good to call this constant POCKET_IC_SERVER_VERSION

done.

michael-weigelt (approved):

PR #348, opened 2024-12-11
fix: remove `rand` dependency from Bitcoin canister

maksymar:
This PR removes the dependency on the `rand` crate from the Bitcoin canister, preparing for the upgrade of the `bitcoin` crate to support `Testnet4`.

The new dependencies rely on the `getrandom` crate (version 0.2.0), which, when compiled for `wasm32-unknown-unknown`, assumes a JavaScript+Node.js environment and imports unsupported modules in the IC production environment.

As a workaround, a lightweight random number generator `SimpleRng` has been implemented. While it is not cryptographically secure, it is sufficient for the `ic-btc-test-utils` library, which is used only for testing purposes (e.g., e2e-tests, canbench).

Changes:
- Added a simple random number generator: `SimpleRng`.
- Removed dependency on the `rand` crate.
- Deduplicated `random_p2*_address` test utilities by moving them to the `ic-btc-test-utils` crate.
- Cleanup: Consolidated `use crate::*` declarations within each file.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3430,7 +3430,6 @@ version = "0.22.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "295642060261c80709ac034f52fca8e5a9fa2c7d341ded5cdb164b7c33768b2a"
 dependencies = [
- "rand 0.6.5",
  "secp256k1-sys",
  "serde",
 ]
--- a/canister/Cargo.toml
+++ b/canister/Cargo.toml
@@ -35,7 +35,7 @@ bench = false
 [dev-dependencies]
 assert_matches = { workspace = true }
 async-std = { version = "1.12.0", features = ["attributes"] }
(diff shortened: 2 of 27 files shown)


dsarlis:
Did you consider using `raw_rand` instead to get a seed? I realize it might not be worth it if this

> While it is not cryptographically secure, it is sufficient for the ic-btc-test-utils library, which is used only for testing purposes (e.g., e2e-tests, canbench).

is really the only use case.

maksymar:
> Did you consider using `raw_rand` management canister call instead to get a seed? 

Yes I did and decided not to, because 

1. it will require a lot of changes -- `raw_rand` is an async call and it also needs the whole environment with management canister (eg. state_machine_tests), while most of the tests are simple unit tests and will have to be re-written to adopt new approach.

2. it does not have to be cryptographically secure since it's going to be used for testing purposes only (generating blockchain with some addresses).

dsarlis (approved):

PR #349, opened 2024-12-11
feat: upgrade bitcoin crate to 0.32.4 for testnet4 support

maksymar:
This PR updates `bitcoin` crate to 0.32.4 to support `Testnet4`.

Changes:
- Upgrade `bitcoin` crate from 0.28.1 to 0.32.4
  - Apply types changes, eg. `Target`, `Value`, difficulty type size etc
- Reuse a `testnet` variant for `testnet4` in `network` enum of `canister.did`
  - Map bitcoin canister `Testnet` to bitcoin crate `Testnet4`
  - Turn off timestamp validation for testnet4 since first 2 blocks are 2 days apart
  - Reset canbench results since testnet reference data was updated
- Upgrade `bitcoin_core` for downloading bitcoin data from 22.0 to 28.0
  - Add `testnet4` network for downloading corresponding data
  - Update `testnet_10k_blocks` test to work with `testnet4` blocks
 
EXC-1813, EXC-1814


validation/src/header.rs
```diff
@@ -270,52 +274,35 @@ fn compute_next_difficulty(
...
-    // Adjusting the newly computed difficulty target so that it doesn't exceed the
-    // max_difficulty_target limit
-    target = Uint256::min(target, max_target(network));
-
-    // Converting the target (Uint256) into a 32 bit representation used by Bitcoin
-    Header::compact_target_from_u256(&target)
+    // the timespan can be negative on testnet networks.
+    let timespan = prev_header.time.saturating_sub(last_adjustment_time) as u64;
+
+    CompactTarget::from_next_work_required(prev_header.bits, timespan, Params::new(*network))
```
maksymar, line 287: Here's link to implementation details:
https://docs.rs/bitcoin/0.32.4/src/bitcoin/pow.rs.html#390-414

canbench_results.yml
```diff
@@ -1,44 +1,44 @@
...
+      instructions: 85751961
       heap_increase: 0
       stable_memory_increase: 0
     scopes: {}
   insert_300_blocks:
     total:
-      instructions: 563254667
-      heap_increase: 6
+      instructions: 561452021
+      heap_increase: 7
```
maksymar, line 11: There's regression from 6 to 7 pages.

canbench_results.yml
```diff
@@ -1,44 +1,44 @@
...
       stable_memory_increase: 0
     scopes: {}
   insert_block_headers:
     total:
-      instructions: 3899095256
-      heap_increase: 2
+      instructions: 3898822288
+      heap_increase: 1
```
maksymar, line 17: Here's an improvement from 2 to 1 page.

validation/src/header.rs
dsarlis: It looks ok to me but I am also not the most familiar with the difficulty calculation anyway. Perhaps @THLO can also confirm here? We do have tests that I trust would have let us know if something was horribly off but an extra pair of eyes shouldn't hurt.

bootstrap/state-builder/src/build_balances.rs
```diff
@@ -59,13 +59,15 @@ fn main() {
 
         // Load the address. The UTXO dump tool we use doesn't output all the addresses
         // we support, so if parsing the address itself fails, we try parsing the script directly.
+        let bitcoin_newtork = into_bitcoin_network(args.network);
```
THLO, line 62: Same typo here.   :slightly_smiling_face:

maksymar: done for all the occurrences.

canister/candid.did
```diff
@@ -1,6 +1,7 @@
 type network = variant {
   mainnet;
   testnet;
+  testnet4;
```
THLO, line 4: I'm not sure adding this field is the right approach.

What if the Bitcoin testnet is (finally) upgraded to version 4? It would then be weird to have a `testnet` variant that is unused.
Of course, the downside is that people may be confused if we simply call the new version `testnet`.
I'm leaning towards option #2 anyway because the current testnet is useless for many people and so I'm hoping that there will be an upgrade.
What do you think?

maksymar: So you're proposing to keep only 2 variants `mainnet` and `testnet` but to use `testnet4` under the hood?
That's an interesting idea, and I agree it'd would reduce the confusion (for not having several testnets).

Let me try to apply this idea into code.

test-utils/src/lib.rs
```diff
@@ -88,14 +93,18 @@ impl BlockBuilder {
...
+                .map(|tx| *tx.compute_txid().as_raw_hash())
+                .clone(),
+        )
+        .unwrap();
+        let merkle_root = TxMerkleNode::from_raw_hash(merkle_root);
 
         let header = match self.prev_header {
-            Some(prev_header) => header(&prev_header, merkle_root),
             None => genesis(merkle_root),
+            Some(prev_header) => header(&prev_header, merkle_root),
```
THLO, line 107: Out of curiosity, why did this line move down?

test-utils/src/lib.rs
```diff
@@ -105,11 +114,14 @@ impl BlockBuilder {
 /// Builds a random chain with the given number of block and transactions
 /// starting with the Regtest genesis block.
 pub fn build_regtest_chain(num_blocks: u32, num_transactions_per_block: u32) -> Vec<Block> {
-    let genesis_block = Block::new(genesis_block(Network::Regtest));
+    let bitcoin_newtork = Network::Regtest;
```
THLO: The same typo!

test-utils/src/lib.rs
maksymar: I thought it's easier to read when the first line is the logical start of the blockchain -- genesis, and all the rest is a header calculation based on a prev_header.

test-utils/src/lib.rs
maksymar: fixed

validation/src/header.rs
THLO: It looks good to me.
We are now using the `from_next_work_required` function from the `bitcoin` crate, right? This function should be implemented correctly.

canister/candid.did
THLO: It can also cause confusion: If a developer makes a test transaction and then tries to find it on a Bitcoin testnet (not testnet4) explorer, he/she won't find it.

This risk could be mitigated by documenting everywhere very clearly that `testnet4` is used.

validation/src/header.rs
maksymar: > We are now using the from_next_work_required function from the bitcoin crate, right?

yes. I'm closing this thread due to LGTMs.

canister/candid.did
maksymar: I have updated the code to re-use `testnet` variant and using `testnet4` under the hood. PTAL

bootstrap/README.md
```diff
@@ -32,7 +32,7 @@ go install github.com/in3rsha/bitcoin-utxo-dump@5723696e694ebbfe52687f51e7fc0ce6
 
 ```shell
 BITCOIN_DIR=/path/to/bitcoin-22.0/
-NETWORK=<mainnet or testnet>
+NETWORK=<mainnet or testnet4>
```
THLO, line 35: Should it still say `testnet4` here?

THLO (approved):

bootstrap/README.md
maksymar: This gets propagated to bitcoin_d so I thought we should not map that.

canbench_results.yml
```diff
@@ -1,44 +1,44 @@
...
+      instructions: 86043839
       heap_increase: 0
       stable_memory_increase: 0
     scopes: {}
   insert_300_blocks:
     total:
-      instructions: 563254667
-      heap_increase: 6
+      instructions: 533444042
+      heap_increase: 10
```
maksymar, line 11: Here's a regression from 6 to 10 pages.

PR #351, opened 2024-12-17
chore: add instructions for generating testnet_blocks.txt

maksymar:
This PR adds instructions for generating `testnet_blocks.txt` hex file.

EXC-1813


dsarlis (commented):
Would it make more sense to guard the logic of saving to a file behind a rust feature? So that we don't need to make source code changes.

maksymar:
> Would it make more sense to guard the logic of saving to a file behind a rust feature? So that we don't need to make source code changes.

thank you, done!

dsarlis (approved):

PR #361, opened 2025-02-05
feat: add get_successors_request_interval histogram metric

maksymar:
This PR adds a histogram metric for tracking `get_successors` request interval.

image

### DIFF
--- a/canister/src/runtime.rs
+++ b/canister/src/runtime.rs
@@ -201,7 +201,7 @@ pub fn get_cycles_balance() -> u64 {
     CYCLES_BALANCE.with(|c| *c.borrow())
 }
 
-pub fn time_secs() -> u64 {
+pub fn time() -> u64 {
     time_nanos() / 1_000_000_000
 }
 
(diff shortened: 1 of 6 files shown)


canister/src/runtime.rs
```diff
@@ -201,21 +201,25 @@ pub fn get_cycles_balance() -> u64 {
...
-/// Returns the current time in seconds.
-#[cfg(target_arch = "wasm32")]
 pub fn time() -> u64 {
     // to get seconds from nanoseconds
-    ic_cdk::api::time() / 1_000_000_000
+    time_nanos() / 1_000_000_000
+}
+
+/// Returns the current time in seconds.
+#[cfg(target_arch = "wasm32")]
```
alexandru-uta, line 210: unimportant for now, but just so that we're aware: since we now have wasm64, if there's ever the case that we compile this canister in 64-bit mode, we'll need to be careful here.

alexandru-uta (approved):

canister/src/runtime.rs
maksymar: noted, thank you!

PR #375, opened 2025-03-04
chore: update docs with fee values for deploying bitcoin canister

maksymar:


bootstrap/README.md
```diff
@@ -268,13 +268,45 @@ $ cargo run --example upload -- \
 
 Prepare upgrade arguments
 ```shell
+# https://internetcomputer.org/docs/references/bitcoin-how-it-works#api-fees-and-pricing
+$ TESTNET_FEES="record { 
+  get_utxos_base = 20_000_000 : nat;
+  get_utxos_cycles_per_ten_instructions = 10 : nat;
```
berestovskyy, line 274: Should we also update the https://internetcomputer.org/docs/references/bitcoin-how-it-works#api-fees-and-pricing

bootstrap/README.md
```diff
@@ -268,13 +268,45 @@ $ cargo run --example upload -- \
...
+  get_utxos_maximum = 4_000_000_000 : nat;
```
berestovskyy, line 275: This also called `minimum` in the reference doc and should be updated?

bootstrap/README.md
```diff
@@ -268,13 +268,45 @@ $ cargo run --example upload -- \
...
+  get_block_headers_maximum = 4_000_000_000 : nat;
+}"
+
+$ MAINNET_FEES="record { 
+  get_utxos_base = 50_000_000 : nat;
+  get_utxos_cycles_per_ten_instructions = 10 : nat;
+  get_utxos_maximum = 10_000_000_000 : nat;
+  get_balance = 10_000_000 : nat;
+  get_balance_maximum = 100_000_000 : nat;
+  get_current_fee_percentiles = 10_000_000 : nat;
```
berestovskyy, line 293: nit: would be nice to have the prices in the same order as the reference doc.

berestovskyy (approved):
Looks good, thanks!

bootstrap/README.md
maksymar: It will be updated separately.

bootstrap/README.md
maksymar: `*_maximum` is a maximum amount of cycles that can be charged, which in doc translates to 'Minimum cycles to send with call' that makes more sense from a developer point of view. So I'll leave it as is.

PR #376, opened 2025-03-05
feat: add default fees for mainnet/testnet networks

maksymar:
This PR adds auto-setup for fees if not explicitly provided during initialisation.

Fees values hardcoded depending on config network (mainnet/testnet) and rely on [docs](https://internetcomputer.org/docs/references/bitcoin-how-it-works#bitcoin-testnet).

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -94,7 +94,7 @@ jobs:
       - name: Run Tests
         shell: bash
         run: |
-          cargo test --release --all-targets --workspace --exclude benchmarks
+          cargo test --release --all-targets --workspace --exclude benchmarks --color always -- --color always
         env:
           RUST_BACKTRACE: 1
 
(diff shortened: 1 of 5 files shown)


.github/workflows/workflow.yml
```diff
@@ -5,6 +5,7 @@ env:
   DFX_VERSION: 0.23.0
   POCKET_IC_SERVER_VERSION: 7.0.0
   CARGO_TERM_COLOR: always # Force Cargo to use colors
+  TERM: xterm-256color
```
maksymar, line 8: adds green ok's.
image

canister/src/state.rs
```diff
@@ -83,14 +83,20 @@ impl State {
         let unstable_blocks =
             UnstableBlocks::new(&utxos, stability_threshold, genesis_block, network);
 
+        let fees = match network {
+            Network::Mainnet => Fees::mainnet(),
+            Network::Testnet => Fees::testnet(),
+            _ => Fees::default(),
```
berestovskyy, line 89: nit: here and below I'd match over all the networks.

berestovskyy (approved):

canister/src/state.rs
maksymar: Done. Thank you!

PR #378, opened 2025-03-06
fix: fix memory leak

maksymar:
This PR removes output points data for the obsolete blocks in unstable blocks tree.


berestovskyy (approved):
Great find, thanks!

PR #383, opened 2025-03-11
fix: [EXC-1987] Fix encoding of get_block_headers metrics on Bitcoin canister

dragoljub-duric:
Those metrics were added to the Metrics structure, but that alone was insufficient.
Each metric must be properly encoded in encode_metrics.


dsarlis (approved):
Since we add more metrics there's a "regression" in the `get_metrics` endpoint. We should update the canbench results and then merge.

PR #401, opened 2025-06-16
feat(benchmarks): add mock time

mducroux:
[XC-394](https://dfinity.atlassian.net/browse/XC-394?atlOrigin=eyJpIjoiNjY4NjU1ZDIxYTY3NDRiNThhMmFhZTVkOTI3NjBkNDYiLCJwIjoiaiJ9): This PR adds the possibility to mock time in `canister/src/runtime.rs`. This is especially useful in benchmark where calling `ic_cdk::api::time()` would always return the same value, namely `1620328630000000023` (May 6, 2021). Because this value is earlier than testnet4 genesis, header timestamp validation would always fail since by consensus rules, timestamp in headers must not be further than 2 hours in the future.

### DIFF
--- a/benchmarks/src/main.rs
+++ b/benchmarks/src/main.rs
@@ -27,6 +27,9 @@ fn init() {
                 .collect(),
         );
     });
+
+    // Set a current time to avoid header validation failures due to blocks appearing to be in the future
+    ic_btc_canister::runtime::set_mock_time_secs(1750077112); // June 2025
 }
 
 // Benchmarks inserting the first 300 blocks of the Bitcoin testnet.
--- a/canister/src/runtime.rs
+++ b/canister/src/runtime.rs
@@ -9,7 +9,6 @@ use ic_cdk::api::call::CallResult;
 use ic_cdk::api::call::RejectionCode;
 #[cfg(not(target_arch = "wasm32"))]
 use serde::Deserialize;
-#[cfg(not(target_arch = "wasm32"))]
 use std::cell::RefCell;
 use std::future::Future;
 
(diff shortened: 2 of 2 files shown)


canister/src/runtime.rs
```diff
@@ -209,7 +212,13 @@ pub fn time_secs() -> u64 {
...
+    MOCK_TIME.with(|t| {
+        if let Some(mock_time) = *t.borrow() {
+            mock_time
+        } else {
+            ic_cdk::api::time()
+        }
+    })
 }
 
 /// Gets current timestamp, in nanoseconds since the epoch (1970-01-01).
```
maksymar, line 224: maybe this function can be used with a mock_time feature, so we don't introduce any debug code into prod.

mducroux: Good point, thanks!

benchmarks/src/main.rs
```diff
@@ -27,6 +27,9 @@ fn init() {
                 .collect(),
         );
     });
+
+    // Set a current time to avoid header validation failures due to blocks appearing to be in the future
+    ic_btc_canister::runtime::mock_time::set_mock_time_secs(1750077112); // June 2025
```
maksymar, line 32: NIT: maybe put this magic number into a constant with corresponding comment

```rs
// Set a fixed mock time (approx. June 2025) to avoid "block from the future" errors.
// Calculated as: 55.5 years * 365.25 days/year * 24h * 60m * 60s ≈ 1_750_000_000 seconds since 1970-01-01 UTC.
let june_2025 = (55.5 * 365.25 * 24.0 * 60.0 * 60.0) as u64;
ic_btc_canister::runtime::mock_time::set_mock_time_secs(june_2025);
```

maksymar (approved):
LGTM given the cleanup comment is fixed.

PR #409, opened 2025-07-09
chore: add canister_id.json and docker build script

mducroux:
This PR:
- adds missing `canister_ids.json` file containing IDs of `ic-btc-canister`and `watchdog` canisters,
- adds `scripts/docker-build` to simplify using docker to reproducibly build canister WASMs.
- breaks down `dfx.json` file to have specific files for test or deployment  purposes
- adds `deployment/mainnet` and `deployment/testnet` folders that contain `dfx.json` and `canister_ids.json` files


canister_ids.json
```diff
@@ -0,0 +1,8 @@
+{
```
maksymar, line 1: I don't think you should add canister_ids.json to the repo, because it really depends on what type of network you're deploying to (mainnet, testnet), therefore the IDs can be different.

can you maybe workaround it in a way that it'll generate a needed canister_ids.json on the fly? and maybe to add it to `.gitignore`?

michael-weigelt (dismissed):

scripts/docker-build
```diff
@@ -0,0 +1,92 @@
+#!/usr/bin/env bash
```
maksymar, line 1: why would one add such a complicated docker build file when the building and extracting instructions pointed [here](https://github.com/dfinity/bitcoin-canister/blob/master/bootstrap/README.md#6-build-canisters) are just 3 commands?

```
# Build all, specifying the path to chunk_hashes.txt
$ docker build --build-arg CHUNK_HASHES_PATH=/bootstrap/chunk_hashes.txt  -t canisters .

# Extract canister's WASM
$ docker run --rm canisters cat /uploader.wasm.gz > uploader.wasm.gz
$ docker run --rm canisters cat /ic-btc-canister.wasm.gz > ic-btc-canister.wasm.gz
```

deployment/mainnet/dfx.json
```diff
@@ -0,0 +1,31 @@
+{
```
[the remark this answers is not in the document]
maksymar: UPD: it's a NIT, it'd be nice not to have those copies of `dfx.json` but it seems there's no way not to have them for e2e-tests. so it's fine.

scripts/docker-build
maksymar: ok, got it, thanks for the context.
good idea putting IDs and configs into dedicated `deployment/<network>` folders.

maksymar (approved):

PR #411, opened 2025-07-16
ci: upload persisted canbench_results.yml file

maksymar:
This PR uploads persisted `canbench_results.yml` files as CI artifacts.

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -18,7 +18,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ ubuntu-latest, macos-latest ]
+        os: [ubuntu-latest, macos-latest]
 
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


e2e-tests/upgradability.sh
```diff
@@ -31,7 +31,16 @@ get_latest_release_url() {
 download_latest_release() {
   local url
   url=$(get_latest_release_url)
+  if [[ -z "$url" ]]; then
+    echo "ERROR: Failed to retrieve latest release URL"
+    exit 1
+  fi
+  echo "Downloading from: $url"
   wget -O "${REFERENCE_CANISTER_NAME}.wasm.gz" "${url}"
+  if [[ ! -f "${REFERENCE_CANISTER_NAME}.wasm.gz" ]]; then
```
berestovskyy, line 40: nit: I'd expect the `wget` to fail in this case because of the `set -e` at the beginning.

berestovskyy (approved):

e2e-tests/upgradability.sh
maksymar: ack.

PR #419, opened 2025-09-19
refactor: validation of bitcoin headers and blocks

gregorydemay:
Unify validation of bitcoin headers and blocks to match the verification done in the [adapter](https://github.com/dfinity/ic/tree/88193c399d792854f98076e5311ffa12726f73e2/rs/bitcoin/adapter). The goal is to have a single implementation for both components.


validation/src/header/tests.rs
```diff
@@ -0,0 +1,402 @@
...
+        MOCK_CURRENT_TIME + 2 * ONE_HOUR - Duration::from_secs(5),
+        MOCK_CURRENT_TIME
+    )
+    .is_ok());
+
+    // 'timestamp_is_less_than_2h_in_future' should return false
+    // because the time is more than 2 hours from the current time.
+    assert_eq!(
+        timestamp_is_less_than_2h_in_future(
+            MOCK_CURRENT_TIME + 2 * ONE_HOUR + Duration::from_secs(10),
```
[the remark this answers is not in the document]
gregorydemay: The method `timestamp_is_less_than_2h_in_future` only returns an error if the block time is strictly more than 2h in the future. So while the offset does not need to be 10s (is like this from before) it needs to be something non-zero.

benchmarks/src/main.rs
```diff
@@ -175,4 +217,59 @@ fn pre_upgrade_with_many_unstable_blocks() -> BenchResult {
...
 }
 
+/// Create a chain of 2 blocks after genesis.
+///
+/// 1st block:
+/// * 1 coinbase transaction with `tx_cardinality` outputs
+///
+/// 2nd block:
+/// * `tx_cardinality` transactions consuming the previous outputs
+fn mini_chain(tx_cardinality: u32) -> [Block; 2] {
```
gregorydemay, line 227: @THLO : As discussed I added 2 benchmarks, 1 for inserting an unstable block with 1k transactions and another one with 10k transactions. How realistic are the blocks produced by the helper `mini_chain`? I took that setup from another pre-existing test ([atomicity_while_ingesting_block](https://github.com/dfinity/bitcoin-canister/blob/145708ff33575cd6a3654c700d43e3d0bf41032e/canister/src/utxo_set.rs#L999)). My impression is that we should do key derivation to have more than 2 addresses involved since addresses are also stored in `OutPointsCache`. WDYT?

validation/src/block/mod.rs
```diff
@@ -0,0 +1,85 @@
...
+        validate_block(block).and_then(|()| {
+            self.header_validator
+                .validate_header(&block.header, current_time)
+                .map_err(ValidateBlockError::InvalidBlockHeader)
+        })
+    }
+}
+
+fn validate_block(block: &bitcoin::Block) -> Result<(), ValidateBlockError> {
+    //check block like in
```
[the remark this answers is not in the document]
gregorydemay: Done.

canister/src/state.rs
```diff
@@ -125,13 +127,12 @@ impl State {
...
-            .map_err(|_| InsertBlockError::PrevHeaderNotFound)?,
-        block.header(),
-        time_secs(),
-    )?;
+    let validator = BlockValidator::new(
+        ValidationContext::new(state, block.header())
+            .map_err(|_| ValidateHeaderError::PrevHeaderNotFound)?,
+        into_bitcoin_network(state.network()),
+    );
+    validator.validate_block(block.internal_bitcoin_block(), duration_since_epoch())?;
```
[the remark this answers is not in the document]
gregorydemay: Discussed offline with @ielashi. The impact for the whole benchmark `pre_upgrade_with_many_unstable_blocks` is not relevant and `get_metrics` is not critical, so we accept those regressions. However, it's currently unclear why those benchmarks were impacted in the first place and will be investigated by Islam in [XC-499](https://dfinity.atlassian.net/browse/XC-499).

benchmarks/src/main.rs
```diff
@@ -154,6 +156,46 @@ fn insert_block_headers_multiple_times() -> BenchResult {
...
+
+    bench_fn(|| {
+        with_state_mut(|s| {
+            ic_btc_canister::state::insert_block(s, block_2).unwrap();
+        });
+    })
+}
+
+#[bench(raw)]
+fn insert_block_with_1k_transactions() -> BenchResult {
```
THLO, line 180: Nit: Is there a reason why the code is duplicated as opposed to introducing a function with the number of transactions as a parameter?

benchmarks/src/main.rs
THLO: I think the Merkle root hash computation is more involved that storing the addresses, so I'd be surprised if using many different addresses made a huge difference.

THLO (approved):

benchmarks/src/main.rs
gregorydemay: No particular reason other than my own lazyness 😄 .  Good catch, refactored.

PR #506, opened 2026-03-20
feat: cherry pick `dfinity/dogecoin-canister@c947b5c` cache unstable blocks in stable memory

mducroux:
Cherry-pick https://github.com/dfinity/dogecoin-canister/commit/c947b5c7be61c1b860a8a4cdf5fdd6a5054c61b3 from Dogecoin canister repository to store unstable blocks in stable memory:

In order to avoid taking up too much heap space, it is desirable to use stable memory as a cache to store unstable blocks. When blocks are removed from the unstable blocks structure, they are also removed from the cache.

A new CachedBlock struct is introduced to replace the Block type at most places. It keeps the block hash while leaving the actual block in cache. It also keeps a copy of pre-calculated difficulty because the difficulty is frequently needed to calculate the longest chain.

The state types are also made generic for the purpose of writing tests to ensure upgrading from old state to new state can succeed. This modification can be reverted once we upgrade all deployed canisters.


ninegua (approved):
LGTM. Thanks!

PR #516, opened 2026-04-15
refactor: move block metrics to `CachedBlock`

mducroux:
## Summary
- Move `fee_rates` and `utxo_delta` from a side `BTreeMap<BlockHash, BlockMetrics>` on `GenericUnstableBlocks` into `CachedBlock` fields, co-locating per-block metrics with the block they belong to.
- Callers that already hold a `&CachedBlock` reference now access metrics directly (`block.fee_rates()`, `block.utxo_delta()`) instead of doing a hash-based lookup through `UnstableBlocks`.
- Metrics are not serialized — they default to `None`/`0` after deserialization, preserving backwards compatibility and the existing fallback recomputation path.
- Remove the `block_fee_rates_are_cleaned_up_on_stable_block_ingestion` test, which became tautological since metrics are now automatically dropped when a block is popped from the tree.

## Test plan
- [x] All existing tests pass (202 unit + 3 main + 1 integration)
- [ ] Verify no benchmark regressions on `get_blockchain_info` and `bitcoin_get_current_fee_percentiles`

### DIFF
--- a/canbench_results.yml
+++ b/canbench_results.yml
@@ -2,95 +2,95 @@ benches:
   bitcoin_get_balance_baseline:
     total:
       calls: 1
-      instructions: 6189268
+      instructions: 6189038
       heap_increase: 0
       stable_memory_increase: 0
     scopes: {}
   bitcoin_get_balance_stress:
     total:
       calls: 1
-      instructions: 254312139
+      instructions: 254311743
       heap_increase: 0
       stable_memory_increase: 0
     scopes: {}
   bitcoin_get_block_headers_baseline:
     total:
       calls: 1
-      instructions: 372085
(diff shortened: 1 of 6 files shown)


ninegua (approved):
Thanks! LGTM. But do you know why this improved the benchmarks?

mducroux:
The improvements we see for `get_blockchain_info` and `bitcoin_get_current_fee_percentiles` are because previously, we had loops calling `unstable_blocks.get_net_utxo_delta(block.block_hash())` per block, which did a `BTreeMap::get`. Now it's `block.utxo_delta()`, a direct field on the `&CachedBlock` that the caller already holds (same for `block.fee_rates()`).

PR #521, opened 2026-04-17
fix: prevent main chain from decreasing when tip is contested

mducroux:
When two competing blocks appeared at the same height with equal accumulated difficulty and equal depth, the main chain selection algorithm truncated to the fork point (the "contested tip" behavior). This caused:
- the reported main chain height to decrease by 1
- `get_utxos` and `get_balance` to temporarily exclude transactions from both competing blocks

On mainnet, all blocks within a difficulty adjustment period share the same target difficulty, so every competing tip triggered this behavior. This is not rare — it happens roughly every few weeks when two miners find a block at approximately the same time.

Bitcoin Core handles ties by staying on whichever chain it was already following. The tip never regresses to the fork point.

The fix in this PR is as follows: remove the contested flag from `main_chain_by_difficulty_inner` and `main_chain_length_by_difficulty_inner`. When two children tie on `(accumulated_difficulty, depth)`, the first-received child wins.

This works because children are ordered by insertion time (`extend` appends via `push`), and using strict `>` in the comparison keeps the first child that set `best_key`. The tiebreaker only applies while branches are exactly equal — as soon as one branch pulls ahead on either difficulty or depth, it wins outright.

### DIFF
--- a/canister/src/api/get_utxos.rs
+++ b/canister/src/api/get_utxos.rs
@@ -824,12 +824,12 @@ mod test {
         );
 
         // Extend block 0 with block 1 that spends the 1000 satoshis and gives them to address 2.
-        let tx = TransactionBuilder::new()
+        let tx_to_address_2 = TransactionBuilder::new()
             .with_input(ic_btc_types::OutPoint::new(coinbase_tx.txid(), 0))
             .with_output(&address_2, 1000)
             .build();
         let block_1 = BlockBuilder::with_prev_header(block_0.header())
-            .with_transaction(tx.clone())
+            .with_transaction(tx_to_address_2.clone())
             .build();
 
         with_state_mut(|state| {
@@ -846,7 +846,7 @@ mod test {
             GetUtxosResponse {
                 utxos: vec![Utxo {
                     outpoint: OutPoint {
-                        txid: tx.txid().into(),
+                        txid: tx_to_address_2.txid().into(),
                         vout: 0,
                     },
                     value: 1000,
@@ -886,8 +886,9 @@ mod test {
             state::insert_block(state, block_1_prime.clone()).unwrap();
         });
 
-        // Because block 1 and block 1' contest with each other, neither of them are included
-        // in the UTXOs. Only the UTXOs of block 0 are returned.
+        // Block 1 was received first, so it wins the tie with block 1'.
+        // The main chain is genesis -> block_0 -> block_1.
+        // address 2 still has the UTXO from block_1.
         assert_eq!(
             get_utxos(GetUtxosRequest {
                 address: address_2.to_string(),
@@ -895,9 +896,16 @@ mod test {
             })
             .unwrap(),
             GetUtxosResponse {
-                utxos: vec![],
-                tip_block_hash: block_0.block_hash().to_vec(),
-                tip_height: 1,
+                utxos: vec![Utxo {
+                    outpoint: OutPoint {
+                        txid: tx_to_address_2.txid().into(),
+                        vout: 0,
+                    },
+                    value: 1000,
+                    height: 2,
+                }],
+                tip_block_hash: block_1.block_hash().to_vec(),
+                tip_height: 2,
                 next_page: None,
             }
         );
@@ -909,8 +917,8 @@ mod test {
             .unwrap(),
             GetUtxosResponse {
                 utxos: vec![],
-                tip_block_hash: block_0.block_hash().to_vec(),
-                tip_height: 1,
+                tip_block_hash: block_1.block_hash().to_vec(),
+                tip_height: 2,
                 next_page: None,
             }
         );
@@ -920,7 +928,12 @@ mod test {
                 filter: None,
             })
             .unwrap(),
-            block_0_utxos
+            GetUtxosResponse {
+                utxos: vec![],
+                tip_block_hash: block_1.block_hash().to_vec(),
+                tip_height: 2,
+                next_page: None,
+            }
         );
 
         // Now extend block 1' with another block that transfers the funds to address 4.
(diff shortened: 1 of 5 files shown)


canister/src/blocktree.rs
```diff
@@ -553,34 +543,23 @@ impl<Block: ChainBlock> BlockTree<Block> {
...
 
         let total_difficulty = self_difficulty + best_key.0;
         let total_depth = 1 + best_key.1;
-
-        if contested {
-            (total_difficulty, total_depth, 1)
-        } else {
-            (total_difficulty, total_depth, 1 + best_length)
-        }
+        (total_difficulty, total_depth, 1 + best_length)
```
gregorydemay: Maybe I'm missing something, but isn't it the case in that method that the return value satisfies `depth==main_chain_length`?:

1. They both start at 1.
2. Depth is incremented iff length is incremented.

If that's the case, why can't we use `main_chain_by_difficulty_inner` that already returns the depth?

canister/src/api/get_utxos.rs
gregorydemay: nit: let's define a variable `block_1_utxos` (like `block_0_utxos`) because this is reused L.892.

canister/src/api/get_utxos.rs
```diff
@@ -1428,9 +1441,10 @@ mod test {
             }
         });
 
-        // Because the forks are of equal length, `A`, the root of the fork,
-        // is considered the tip at zero confirmations.
-        assert_tip_at_confirmations(0, chain[0].block_hash());
+        // With the first-received-child-wins rule, the first fork (chain)
+        // is chosen as the main chain. All blocks have stability count >= 0,
+        // so the tip at zero confirmations is F.
```
gregorydemay, line 1436: nit: What is meant with `F` in
```
        // so the tip at zero confirmations is F.
```
?

gregorydemay (approved):
Thanks @mducroux ! besides the previous understanding question LGTM!

gregorydemay:
One subtle behavioral note worth calling out (not blocking).

The PR description and the doc comment on `BlockchainInfo` (`interface/src/lib.rs:812`) equate the new behavior with Bitcoin Core's "stay on whichever chain you were already following." These match in most cases but are not strictly identical: the canister recomputes the main chain from the tree on every call, and the tiebreaker always picks the **first-inserted** child, not the **currently-active** one.

Concrete diverging scenario:
1. Receive `X` — chain follows X.
2. `Y` arrives and overtakes (`Y1`, `Y2`) — chain switches to Y.
3. `X` catches up (`X1`, `X2`) so that the X-branch and Y-branch tie again on `(accumulated_difficulty, depth)`.
4. **Canister**: reverts to X (X is the first-inserted child of the root, so iteration sets `best_key` from X first and the strict `>` keeps it).
   **Bitcoin Core**: stays on Y.

This produces a momentary reorg in the canister that wouldn't happen in Bitcoin Core, and resolves as soon as either side pulls ahead again. It's strictly better than the old behavior — no anchor truncation, height never drops to the fork point — so I'd accept it as-is.

Suggestion: soften the wording from "Bitcoin Core handles ties by staying on whichever chain it was already following" to something like "the first-received branch wins ties" in both the PR description and the `BlockchainInfo` doc comment, since that's what's actually implemented.

canister/src/api/get_utxos.rs
mducroux: Thanks, done here: 2b22be89889f555ef8bf8e478205e38a5e64c1e0

canister/src/api/get_utxos.rs
mducroux: F refers to the last block of the first fork in the ASCII diagram at     
  lines 1414–1416:                                                         
                                                                           
  A -> B -> C -> D -> E -> F                                               
  |                                                                        
   \-> B'-> C'-> D'-> E'-> F'

mducroux:
> These match in most cases but are not strictly identical

That's a good point! However I don't think it is said anywhere that the behaviour is now equivalent to Bitcoin core so I would not make any changes to the doc.

canister/src/blocktree.rs
mducroux: Yes we could reuse the existing `main_chain_by_difficulty_inner` returned depth but the problem is that this function does more than what is needed to get the length. In particular, it allocates a `Vec<&Block>` at every recursive call which is not needed. The point of having two separate inner function was efficiency. To confirm, I measure the performance when reusing the existing `main_chain_by_difficulty_inner` and obtained the following regression: 

| status | name                                | calls |     ins |  ins Δ% | HI |  HI Δ% | SMI |  SMI Δ% |
|--------|-------------------------------------|-------|---------|---------|----|--------|-----|---------|
|   +    | get_blockchain_info_many_branches   |       |   1.16M | +30.73% |  0 |  0.00% |   0 |   0.00% |
|   +    | bitcoin_get_current_fee_percentiles |       |  55.20K | +16.84% |  0 |  0.00% |   0 |   0.00% |
|   +    | get_blockchain_info_with_forks      |       | 466.14K | +14.44% |  0 |  0.00% |   0 |   0.00% |
|   +    | bitcoin_get_block_headers_stress    |       |   3.22M | +12.83% |  0 |  0.00% |   0 |   0.00% |
|   +    | get_blockchain_info_single_chain    |       | 419.99K | +11.51% |  0 |  0.00% |   0 |   0.00% |
|   +    | get_metrics                         |       |   3.77M |  +7.07% |  0 |  0.00% |   0 |   0.00% |
|   +    | bitcoin_get_block_headers_baseline  |       | 386.81K |  +3.91% |  0 |  0.00% |   0 |   0.00% |

So I would suggest to keep it as is.

gregorydemay: Thanks for digging into that! It makes sense to have both functions then. I think it would be useful to explain this in the Rust doc of that method. Something like, 
same logic as in `main_chain_by_difficulty_inner` but avoid unnecessary allocation of `Vec<&Block>`.

mducroux: Good point, done in b9c5776b1518486f24be7130ba1094a98e60e905. I also removed the third element of the triple returned by `main_chain_length_by_difficulty_inner ` which was actually redundant with the second one.

PR #524, opened 2026-05-13
test: DEFI-2100: Migrate scenario-1 e2e test from dfx to PocketIC

mbjorkqvist:
Replace the `dfx`-based `e2e-tests/scenario-1.sh` shell script with a Rust integration test driven by PocketIC.

**Why.** The shell test was slow (full `dfx start` per run), brittle (assertions done by substring-matching Candid text output), and excluded from typed checks. Moving to PocketIC gives us a faster local feedback loop and type-safe interactions with the canister via `ic-btc-interface`.

**What changes.**
  - New `e2e-tests/scenario-1/tests/integration.rs` reproduces every assertion from the deleted shell script: balance & UTXO queries on three addresses, the 1000-UTXO response cap, replicated-mode rejection of `*_query` methods, fee-percentile smoke calls, and byte-exact comparison of all six block headers.
  - Shared address constants moved from the source canister's `main.rs` into a new `e2e-tests/scenario-1/src/lib.rs` so the canister binary and the integration test share one definition.
  - The `e2e-scenario-1` CI job drops `dfx` and `dfinity/setup-dfx`, installs PocketIC, builds the source canister WASM with `scripts/build-canister.sh scenario-1`, and runs `cargo test --release -p scenario-1`. WASM paths are passed via `IC_BTC_CANISTER_WASM_PATH` and `E2E_SCENARIO_1_WASM_PATH` env vars; the `ic-btc-canister` WASM still comes from the `canister-build-reproducibility` artifact.
  - `scenario-1` is excluded from the main `cargo-tests` matrix (it needs the built WASM artifacts, which only the dedicated job has) and from its `needs:` list (no longer dependent on `cargo-build`).
  - The stable-block check reads `stable_height` from the canister's `/metrics` HTTP endpoint, matching the original script's approach rather than blindly ticking. The value is parsed as `f64` so a future encoder change to `"3.0"` doesn't break the test, and both `stable_height N` and `stable_height{...} N` line forms are accepted in case labels are added.
  - Tests run on an application subnet via `PocketIcBuilder::with_application_subnet()` so behaviour reflects production configuration.


e2e-tests/scenario-1/tests/integration.rs
```diff
@@ -0,0 +1,393 @@
...
+    headers: Vec<(String, String)>,
+    body: ByteBuf,
+}
+
+#[derive(CandidType, Deserialize)]
+struct HttpResponse {
+    status_code: u16,
+    headers: Vec<(String, String)>,
+    body: ByteBuf,
+}
```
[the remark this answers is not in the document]
mbjorkqvist: Done in the next commit — replaced the local definitions with `ic_btc_canister::types::{HttpRequest, HttpResponse}` via a dev-dependency on `ic-btc-canister`.

e2e-tests/scenario-1/tests/integration.rs
```diff
@@ -0,0 +1,379 @@
...
+    pic: PocketIc,
+    btc_id: Principal,
+}
+
+impl Setup {
+    fn new() -> Self {
+        let source_wasm = load_wasm("E2E_SCENARIO_1_WASM_PATH", "scenario-1");
+        let btc_wasm = load_wasm("IC_BTC_CANISTER_WASM_PATH", "ic-btc-canister");
+
+        let pic = PocketIcBuilder::new().with_application_subnet().build();
```
gregorydemay: This part also mentioned in the PR description
> Tests run on an application subnet via PocketIcBuilder::with_application_subnet() so behaviour reflects production configuration.

is I think not correct. If possible we should install the bitcoin canister on a system subnet to have a closer setup to what we have in production.

e2e-tests/scenario-1/tests/tests.rs
```diff
@@ -0,0 +1,379 @@
+use candid::{CandidType, Deserialize, Principal};
```
gregorydemay, line 1: nit: why no just name this file tests.rs?

e2e-tests/scenario-1/tests/integration.rs
```diff
@@ -0,0 +1,379 @@
...
+    );
+
+    // Fee percentiles smoke test. The result is intentionally not asserted; these
+    // calls exist only to exercise the endpoint for profiling, matching the
+    // behaviour of the original scenario-1.sh script.
+    let fee_req = || GetCurrentFeePercentilesRequest {
+        network: NetworkInRequest::Regtest,
+    };
+    setup.bitcoin_get_current_fee_percentiles(fee_req());
+    setup.bitcoin_get_current_fee_percentiles(fee_req());
```
gregorydemay: I understand that this was like this in the script but I'm honestly not sure what is meant with "This is only for profiling purposes.".

.github/workflows/workflow.yml
```diff
@@ -167,26 +167,34 @@ jobs:
...
+          rustup default ${RUST_VERSION}
+          rustup target add wasm32-unknown-unknown
+
+      - name: Install PocketIC
+        run: |
+          wget https://github.com/dfinity/pocketic/releases/download/$POCKET_IC_SERVER_VERSION/pocket-ic-x86_64-linux.gz
+          gzip -d pocket-ic-x86_64-linux.gz
+          mv pocket-ic-x86_64-linux pocket-ic
+          chmod +x pocket-ic
+          mv pocket-ic ./canister/
```
gregorydemay: nit: There is a Github action for that. Example [here](https://github.com/dfinity/sol-rpc-canister/blob/5f4671b7944a183f8ebad4b1a37bccac8926f027/.github/workflows/ci.yml#L184)

e2e-tests/scenario-1/tests/integration.rs
```diff
@@ -0,0 +1,379 @@
...
+        let bytes = self
+            .pic
+            .query_call(
+                self.btc_id,
+                Principal::anonymous(),
+                "http_request",
+                candid::encode_one(request).unwrap(),
+            )
+            .ok()?;
+        let response = candid::decode_one::<HttpResponse>(&bytes).ok()?;
```
gregorydemay: why silently ignore errors (`ok()?`)? Shouldn't we always be able to make such a request and decode the response?

e2e-tests/scenario-1/tests/integration.rs
```diff
@@ -0,0 +1,379 @@
...
+        assert_eq!(
+            response.status_code,
+            200,
+            "metrics endpoint returned {}: {}",
+            response.status_code,
+            String::from_utf8_lossy(&response.body)
+        );
+        let body = String::from_utf8(response.body.into_vec()).ok()?;
```
gregorydemay: same here.

gregorydemay (commented):
Thanks a lot @mbjorkqvist for starting this! Just a couple of minor comments but otherwise LGTM!

e2e-tests/scenario-1/tests/integration.rs
mbjorkqvist: Addressed in 1869e1c — bitcoin canister now installed on a `bitcoin` subnet via `with_bitcoin_subnet()`, source canister stays on an application subnet. PR description updated accordingly.

e2e-tests/scenario-1/tests/tests.rs
mbjorkqvist: Done in 4aa4d0c — `git mv integration.rs tests.rs`.

e2e-tests/scenario-1/tests/integration.rs
mbjorkqvist: Dug into the history: the "for profiling purposes" comment and the two unchecked calls came from PR #50 (Sep 2022, Islam — "instrument get_current_fee_percentiles"), which also added two lines to `e2e-tests/instructions_count.txt` recording each call's instruction count (`1552293798` and `1552293676`) as a hand-maintained profiling baseline. The two-call structure exercises the cold-compute path followed by the cache-hit path (`canister/src/api/fee_percentiles.rs:57-61`).

That baseline file has since been deleted, and the endpoint is now covered with far more rigour by the canbench benchmark `bitcoin_get_current_fee_percentiles` (visible in every CI run via `canbench_results.yml`) and by dedicated unit tests in `canister/src/api/fee_percentiles.rs` (including caching behaviour). So the two e2e calls produce no signal today.

Dropped both calls (and the helper) in ade73f9.

.github/workflows/workflow.yml
mbjorkqvist: Good catch — done in 6c35d3f. Pinned to the same SHA as sol-rpc-canister (`20c33db1aa87cc6ece50857ac632c37acf5e0322`). The action exports `POCKET_IC_BIN`, so the cargo test step picks it up automatically and we can drop the `mv pocket-ic ./canister/` step. Scoped to the `e2e-scenario-1` job; the other manual install blocks are pre-existing and can be cleaned up separately.

e2e-tests/scenario-1/tests/integration.rs
mbjorkqvist: Right — `stable_height` is unconditionally encoded as a gauge from the very first `/metrics` request (`canister/src/api/metrics.rs:50-54`), so every step *must* succeed. Fixed in e226323: replaced every `.ok()?` with a precise `.expect(...)` and changed the return type from `Option<u32>` to `u32`. `tick_until_stable_height` simplifies to a direct numeric comparison.

e2e-tests/scenario-1/tests/integration.rs
mbjorkqvist: Fixed in e226323 (see reply above).

e2e-tests/scenario-1/tests/tests.rs
```diff
@@ -20,13 +20,21 @@ impl Setup {
         let source_wasm = load_wasm("E2E_SCENARIO_1_WASM_PATH", "scenario-1");
         let btc_wasm = load_wasm("IC_BTC_CANISTER_WASM_PATH", "ic-btc-canister");
 
-        let pic = PocketIcBuilder::new().with_application_subnet().build();
-
-        let source_id = pic.create_canister();
+        // Install the bitcoin canister on a bitcoin subnet to match production; the
+        // source canister is a test fixture and lives on a separate application subnet.
```
gregorydemay: nit: 
>         // source canister is a test fixture and lives on a separate application subnet.

Basically this mocks the role of the adapter, no? It's therefore questionable to have it on another subnet.

gregorydemay (approved):

e2e-tests/scenario-1/tests/tests.rs
mbjorkqvist: Good point — the source canister mocks the adapter, which is co-located with the bitcoin canister in production, not a separate canister on another subnet. Fixed in 1a60f6a: dropped `.with_application_subnet()`, both canisters now on the bitcoin subnet. PR description updated to match. Scenario-1 test re-run locally and still passes.

PR #525, opened 2026-05-19
test: DEFI-2100: Migrate scenario-2 and scenario-3 e2e tests to PocketIC

mbjorkqvist:
Continue the dfx → PocketIC migration started in #524, this time covering `scenario-2` and `scenario-3`, and extracting the reusable bits into a shared `e2e-test-utils` crate so the remaining migrations are mechanical.

**Why.** The same motivation as #524: shell tests are slow (full `dfx start` per run), brittle (substring-matching Candid text), and excluded from typed checks. Scenario-1 proved the PocketIC pattern; this PR generalises it. After this lands, 7 bitcoin-canister shell scripts remain (set_config, post_upgrade_config, bitcoin-canister-metadata, cycles_burn, disable-api-…, upgradability, charge-cycles-on-reject), all of which can reuse the helpers introduced here.

**What changes.**
  - New `e2e-tests/test-utils/` crate (`e2e-test-utils`) collects the shared infrastructure extracted from scenario-1: `load_wasm`, the bitcoin-subnet PocketIC bootstrap, generic candid call helpers (`query`, `update`, `update_raw`), bitcoin-canister introspection (`get_blockchain_info`, `get_stable_height`, `tick_until_main_chain_height`, `tick_until_stable_height`) and typed wrappers (`bitcoin_get_balance{_query}`, `bitcoin_get_utxos{_query}`, `bitcoin_get_block_headers`, `bitcoin_send_transaction`).
  - `scenario-1/tests/tests.rs` is refactored to consume `e2e-test-utils`. Behaviour is unchanged; the test still passes locally in ~8s. The file shrinks from 376 → 187 lines (the genesis-through-block-5 header blobs now dominate).
  - `scenario-2/tests/tests.rs` reproduces every assertion from the deleted `scenario-2.sh`: blockchain height 4 after the 4 source-canister blocks are ingested, ADDRESS balance of 40_000 satoshi (4 × 10_000 × 1 sat), and the 1000-UTXO response cap. The hard-coded address moves into a new `scenario-2/src/lib.rs` so the canister bin and the test share one definition (same pattern as scenario-1).
  - `scenario-3/tests/tests.rs` covers both branches of `bitcoin_send_transaction`: a valid (empty segwit-encoded) transaction that reaches the source canister via the inter-canister `bitcoin_send_transaction_internal` call (verified by querying the source's `get_last_transaction`), and a malformed-bytes path that must reject with a message containing `MalformedTransaction`. The bitcoin canister's `send_transaction` awaits the inter-canister call before returning, so no extra ticks are needed between the update call and the source-canister query.
  - The `e2e-scenario-2` and `e2e-scenario-3` CI jobs drop `dfx` and `dfinity/setup-dfx`, install PocketIC via the `dfinity/pocketic@20c33db1aa87cc6ece50857ac632c37acf5e0322` action, build the relevant source-canister WASM with `scripts/build-canister.sh`, and run `cargo test --release -p scenario-{2,3}`. WASM paths are passed via `IC_BTC_CANISTER_WASM_PATH` and `E2E_SCENARIO_{2,3}_WASM_PATH`; the `ic-btc-canister` WASM still comes from the `canister-build-reproducibility` artifact. Both jobs lose their `cargo-build` dependency since cargo runs the test directly.
  - `scenario-2` and `scenario-3` are excluded from the main `cargo-tests` matrix (same reason as scenario-1: they need the PocketIC binary and the built WASM artifacts that only the dedicated e2e jobs produce).
  - `e2e-tests/utils.sh` is left in place — still sourced by the 7 remaining shell scripts. It can be deleted after the last shell script is gone.


.github/workflows/workflow.yml
```diff
@@ -95,7 +95,7 @@ jobs:
       - name: Run Tests
         shell: bash
         run: |
-          cargo test --release --all-targets --workspace --exclude benchmarks --exclude e2e-cdk-bitcoin-canister --exclude scenario-1 -- --color always
+          cargo test --release --all-targets --workspace --exclude benchmarks --exclude e2e-cdk-bitcoin-canister --exclude scenario-1 --exclude scenario-2 --exclude scenario-3 -- --color always
```
gregorydemay: nit: could we use a regex to exclude those end--to-end tests?

.github/workflows/workflow.yml
```diff
@@ -153,9 +153,13 @@ jobs:
         env:
           SHELLCHECK_OPTS: -e SC1090 -e SC2119 -e SC1091
 
-  e2e-scenario-1:
+  e2e-scenario:
     runs-on: ubuntu-latest
     needs: canister-build-reproducibility
+    strategy:
+      fail-fast: false
+      matrix:
```
gregorydemay, line 161: nice simplification with the matrix 🥳

e2e-tests/scenario-1/tests/tests.rs
```diff
@@ -239,13 +29,27 @@ fn utxos_req(address: &str) -> GetUtxosRequest {
...
+        btc_wasm,
+    );
 
     // Wait until all 5 blocks have been ingested. The scenario-1 canister serves 7
     // GetSuccessors responses (one per heartbeat call); 500 ticks is a generous ceiling.
-    setup.tick_until_main_chain_height(5, 500);
+    tick_until_main_chain_height(&pic, btc_id, 5, 500);
 
-    let info = setup.get_blockchain_info();
+    let info = get_blockchain_info(&pic, btc_id);
```
gregorydemay, line 52: nit: you could have kept the `Setup` struct in the test-utils libs and use it here. to avoid touching too much this file. This is mostly a stylistic choice (diff is small) so nothing to do, just thought worth mentioning.

e2e-tests/dfx.json
```diff
@@ -18,20 +18,6 @@
       "build": "../scripts/build-canister.sh scenario-1",
       "wasm": "../target/wasm32-unknown-unknown/release/scenario-1.wasm.gz"
     },
```
gregorydemay, line 20: nit: shouldn't we also remove that block?

gregorydemay (approved):
Thanks @mbjorkqvist !

e2e-tests/scenario-1/tests/tests.rs
mbjorkqvist: Good point! I think this would make sense, especially with more scenario migrations coming up. I'll do this refactor in a follow-up PR before continuing with the next dfx-to-pocketIC migration.

e2e-tests/dfx.json
mbjorkqvist: Good catch — but this one I'd like to leave for the next migration step. The `e2e-scenario-1` dfx entry is still load-bearing for `e2e-tests/charge-cycles-on-reject.sh`, which deploys it as the blocks source (`dfx deploy e2e-scenario-1` and `dfx canister id e2e-scenario-1`). The scenario-2/3 entries removed in f8860c6 had no remaining shell-script consumers, which is why those were safe to drop. I'll remove the `e2e-scenario-1` block once `charge-cycles-on-reject.sh` is migrated to PocketIC (one of the 7 remaining shell tests).

.github/workflows/workflow.yml
mbjorkqvist: Good point, changed to `--exclude 'scenario-*'`.

PR #526, opened 2026-05-21
refactor: DEFI-2100: Setup struct to bundle PocketIC e2e test boilerplate

mbjorkqvist:
Every e2e test currently repeats the same ~5-line bootstrap (load source WASM, load bitcoin canister WASM, build a PocketIC with a bitcoin subnet, install the source canister, install the bitcoin canister with blocks_source wired to the source) and threads `&pic, btc_id` through every bitcoin canister call. With three callers in place — and seven more shell tests still to migrate — the per-test boilerplate is worth extracting.

Setup owns the PocketIC instance plus the two canister ids and exposes thin method wrappers around the bitcoin canister surface (`get_blockchain_info`, `tick_until_*`, `bitcoin_get_balance{,_query}`, `bitcoin_get_utxos{,_query}`, `bitcoin_get_block_headers`, `bitcoin_send_transaction`). Fields are public so tests can still reach the underlying PocketIc for source-canister queries (scenario-3) and raw update calls that inspect the reject response (scenario-1 / scenario-3). The free functions stay exported and now delegate to the same code paths the methods call into.

scenario-1's test file shrinks from 199 → 175 lines, scenario-2 from 60 → 43, scenario-3 from 59 → 47, and the per-test InitConfig drops the blocks_source override since Setup::new wires it. No behavioural change.

### DIFF
--- a/e2e-tests/test-utils/src/lib.rs
+++ b/e2e-tests/test-utils/src/lib.rs
@@ -299,15 +299,6 @@ pub fn bitcoin_send_transaction(pic: &PocketIc, btc_id: Principal, req: SendTran
 
 // ---------- Setup ----------
 
-/// Bundles a PocketIC instance with a deployed bitcoin canister and its
-/// source canister, plus thin method wrappers around the most common
-/// bitcoin canister calls.
-///
-/// Fields are public so tests can still reach the underlying `PocketIc`
-/// and canister ids for things that don't fit the bitcoin-canister method
-/// surface — source-canister queries (e.g. scenario-3's
-/// `get_last_transaction`), and raw update calls that inspect the reject
-/// response (e.g. asserting on `MalformedTransaction`).
 pub struct Setup {
     pub pic: PocketIc,
     pub btc_id: Principal,
--- a/e2e-tests/test-utils/src/lib.rs
+++ b/e2e-tests/test-utils/src/lib.rs
@@ -317,8 +317,16 @@ pub struct Setup {
 impl Setup {
     /// Brings up a fresh PocketIC instance with a bitcoin subnet, installs
     /// the source canister, then installs the bitcoin canister with
-    /// `init.blocks_source` overridden to point at the source canister id.
+    /// `init.blocks_source` wired to the source canister id.
+    ///
+    /// Panics if `init.blocks_source` is `Some(_)`: the source canister id
+    /// is created inside this constructor, so a caller-provided value can
+    /// only be a mistake.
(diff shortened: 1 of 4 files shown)


e2e-tests/test-utils/src/lib.rs
```diff
@@ -296,3 +296,85 @@ pub fn bitcoin_get_block_headers(
...
+    /// `init.blocks_source` overridden to point at the source canister id.
+    pub fn new(source_wasm_env: &str, source_name: &str, init: InitConfig) -> Self {
+        let source_wasm = load_wasm(source_wasm_env, source_name);
+        let btc_wasm = load_wasm("IC_BTC_CANISTER_WASM_PATH", "ic-btc-canister");
+        let (pic, bitcoin_subnet) = pocket_ic_with_bitcoin_subnet();
+        let source_id = install_canister_on_subnet(&pic, bitcoin_subnet, source_wasm, vec![]);
+        let init = InitConfig {
+            blocks_source: Some(source_id),
+            ..init
+        };
```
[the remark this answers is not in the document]
mbjorkqvist: Good catch. Tightened this with an `assert!(init.blocks_source.is_none(), ...)` and a doc note explaining the precondition. The source canister id is created inside `Setup::new`, so there's no meaningful value a caller could pass — failing loudly on misuse is preferable to a silent override. A `Result` would force every test to `.unwrap()` an error that can't realistically arise, and an explicit `blocks_source` parameter doesn't fit the constructor's shape.

e2e-tests/test-utils/src/lib.rs
```diff
@@ -296,3 +296,93 @@ pub fn bitcoin_get_block_headers(
...
+
+/// Bundles a PocketIC instance with a deployed bitcoin canister and its
+/// source canister, plus thin method wrappers around the most common
+/// bitcoin canister calls.
+///
+/// Fields are public so tests can still reach the underlying `PocketIc`
+/// and canister ids for things that don't fit the bitcoin-canister method
+/// surface — source-canister queries (e.g. scenario-3's
+/// `get_last_transaction`), and raw update calls that inspect the reject
+/// response (e.g. asserting on `MalformedTransaction`).
```
gregorydemay: nit: remove? This is test infrstructure so I would keep Rust doc to a minimum.

gregorydemay (approved):
Thanks @mbjorkqvist !

e2e-tests/test-utils/src/lib.rs
mbjorkqvist: Done — dropped the docstring on the `Setup` struct.

PR #533, opened 2026-05-28
chore: proposal to upgrade the Bitcoin canister to release/2026-05-27

mbjorkqvist:
This upgrade changes the main chain selection algorithm so that the reported main chain no longer shortens when two competing blocks appear at the same height with equal accumulated difficulty and equal depth (a contested tip). Previously, the algorithm truncated to the fork point in this case, decreasing the reported main chain height by one and causing `bitcoin_get_utxos` and `bitcoin_get_balance` to temporarily exclude transactions from both competing blocks. On mainnet, all blocks within a difficulty adjustment period share the same target difficulty, so any competing tip triggered this behavior, which occurs roughly every few weeks when two miners find a block at approximately the same time. With this change, ties on `(accumulated_difficulty, depth)` are broken by keeping the first-received child, matching Bitcoin Core's behavior of staying on whichever chain the node was already following. The tiebreaker only applies while branches are exactly equal — as soon as one branch pulls ahead on either difficulty or depth, it wins outright.

### DIFF
--- a/deployment/mainnet/bitcoin_canister_upgrade_2026_05_27.md
+++ b/deployment/mainnet/bitcoin_canister_upgrade_2026_05_27.md
@@ -1,0 +1,48 @@
+# Proposal to upgrade the Bitcoin canister
+
+Repository: `https://github.com/dfinity/bitcoin-canister.git`
+
+Git hash: `60a9338940034536957d50a85444419501b1ecc7`
+
+New compressed Wasm hash: `99ed101afe538f6c08134bb2cfda7ab7b3000a5cc44c65c3bcbf1e5772803f90`
+
+Upgrade args hash: `1c08f4660179db8e87a4a5060ca93ea2f476fed7c25a2bd92f9cf8375eff87d9`
+
+Target canister: `ghsi2-tqaaa-aaaan-aaaca-cai`
+
+Previous Bitcoin proposal: https://dashboard.internetcomputer.org/proposal/141435
+
+---
+
+## Motivation
+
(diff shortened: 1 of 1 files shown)


deployment/mainnet/bitcoin_canister_upgrade_2026_05_27.md
```diff
@@ -0,0 +1,48 @@
+# Proposal to upgrade the Bitcoin canister
+
+Repository: `https://github.com/dfinity/bitcoin-canister.git`
+
+Git hash: `60a9338940034536957d50a85444419501b1ecc7`
+
+New compressed Wasm hash: `99ed101afe538f6c08134bb2cfda7ab7b3000a5cc44c65c3bcbf1e5772803f90`
+
+Upgrade args hash: `1c08f4660179db8e87a4a5060ca93ea2f476fed7c25a2bd92f9cf8375eff87d9`
```
gregorydemay, line 9: verified 🎉

deployment/mainnet/bitcoin_canister_upgrade_2026_05_27.md
```diff
@@ -0,0 +1,48 @@
+# Proposal to upgrade the Bitcoin canister
+
+Repository: `https://github.com/dfinity/bitcoin-canister.git`
+
+Git hash: `60a9338940034536957d50a85444419501b1ecc7`
+
+New compressed Wasm hash: `99ed101afe538f6c08134bb2cfda7ab7b3000a5cc44c65c3bcbf1e5772803f90`
```
gregorydemay, line 7: verified 🎉

gregorydemay (approved):
Thanks @mbjorkqvist !

PR #534, opened 2026-06-10
test: DEFI-2100: Migrate bitcoin-canister-metadata e2e test to PocketIC

mbjorkqvist:
## Summary
Replaces the dfx-based `bitcoin-canister-metadata.sh` e2e test with a Rust integration test under `e2e-tests/bitcoin-canister-metadata/`, continuing the DEFI-2100 migration of e2e tests off dfx and onto PocketIC.

Since PocketIC 11 exposes no API to read a canister's metadata custom sections — and dfx was only ever used here to read that section — the test reads the `icp:public candid:service` section directly from the bitcoin canister wasm (gunzipping the CI artifact when needed) and asserts it matches `canister/candid.did`. This preserves the same build-artifact coverage as the original shell test, without the dfx round-trip.

The test is folded into the `e2e-scenario` CI matrix as a sourceless entry; the standalone `bitcoin_canister_metadata` workflow job is dropped and the crate is excluded from the workspace `cargo test` run since it needs the prebuilt wasm.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


gregorydemay (approved):
Thanks @mbjorkqvist !

PR #535, opened 2026-06-10
test: DEFI-2100: Migrate cycles_burn e2e test to PocketIC

mbjorkqvist:
## Summary
Replaces the dfx-based `cycles_burn.sh` e2e test with a Rust integration test under `e2e-tests/cycles-burn/` driven by PocketIC, continuing the DEFI-2100 migration of e2e tests off dfx and onto PocketIC.

The original deployed the bitcoin canister with cycles and `burn_cycles = enabled` and checked that its balance dropped to zero. The canister burns its balance in the heartbeat, which under PocketIC already runs as part of installation, so a canister installed with `burn_cycles = enabled` has a zero balance by the time it is up. The test asserts this, and adds a contrasting `burn_cycles = disabled` canister that retains its funded balance — demonstrating the flag is what drives the burn and guarding against a false pass if the canister were ever installed unfunded.

The test is folded into the `e2e-scenario` CI matrix as a sourceless entry; the standalone `cycles_burn` workflow job is dropped and the crate is excluded from the workspace `cargo test` run since it needs the prebuilt bitcoin canister wasm.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


.github/workflows/workflow.yml
```diff
@@ -95,7 +95,7 @@ jobs:
       - name: Run Tests
         shell: bash
         run: |
-          cargo test --release --all-targets --workspace --exclude benchmarks --exclude e2e-cdk-bitcoin-canister --exclude 'scenario-*' --exclude post-upgrade-config --exclude set-config -- --color always
+          cargo test --release --all-targets --workspace --exclude benchmarks --exclude e2e-cdk-bitcoin-canister --exclude 'scenario-*' --exclude post-upgrade-config --exclude set-config --exclude cycles-burn -- --color always
```
gregorydemay, line 98: nit: why not naming it something like `scenario-cycles-burn` to have it included in the previous regex `--exclude 'scenario-*'` (remark also goes for `post-upgrade-config` and `set-config`, but maybe that's for another PR?)

gregorydemay (approved):
Thanks @mbjorkqvist !

.github/workflows/workflow.yml
mbjorkqvist: Good point — I'd rather not single out `cycles-burn` with a one-off `scenario-*` rename, since it's a driver crate like `post-upgrade-config`/`set-config` rather than a `scenario-*` source canister. Instead I've filed DEFI-2893 to rename *all* the e2e test crates to an `e2e-` prefix (`e2e-scenario-1`, `e2e-set-config`, `e2e-cycles-burn`, …), which lets the exclude list collapse to `--exclude 'e2e-*'`. No action in this PR.

https://dfinity.atlassian.net/browse/DEFI-2893

PR #538, opened 2026-07-21
perf: DEFI-2954: Avoid deep-cloning ingesting_block on every heartbeat

mbjorkqvist:
## Purpose

The Bitcoin mainnet-staging canister (`axowo`, subnet `io67a`) stopped making sync progress the moment its replica was upgraded to a build that enables the **deterministic memory tracker (DMT)**. DMT charges instructions for every page access — including reads — which the previous memory tracker did not.

`ingest_stable_blocks_into_utxoset` cloned and `PartialEq`-compared the entire `ingesting_block` on **every heartbeat**, purely to decide whether any ingestion work had happened. That structure is the full block (stored twice internally, no `Arc`) plus the monotonically growing `utxos_delta` — all heap data. Under the old tracker those reads were free; under DMT they are charged per accessed page, against the same instruction counter the ingestion loop uses to time-slice. The block half is constant across all slices of a block yet is re-cloned/compared every heartbeat.

This PR replaces the clone + deep-compare with a cheap boolean derived directly from the slicing results, removing that avoidable per-slice cost from the block-ingestion hot path.

## Notes

- Behaviour is unchanged: a paused slice always mutates `ingesting_block` (via the round counter), so the previous comparison already returned `true` in that case; returning `true` on a paused slice is also the safe choice for the self-imposed instruction-limit guard (the heartbeat then declines to also fetch/process in the same round).
- This is a low-risk first lever. Profiling suggests the clone/compare is a contributor (~10–25% of per-heartbeat page traffic) rather than the dominant cost, so it may not fully resolve the stall on its own; the remaining cost is the UTXO-set operations under DMT.
- Tracking / full analysis: DEFI-2954.

### DIFF
--- a/canister/src/heartbeat.rs
+++ b/canister/src/heartbeat.rs
@@ -4,7 +4,7 @@ use crate::{
     state::{self, ResponseToProcess},
     types::{
         GetSuccessorsCompleteResponse, GetSuccessorsRequest, GetSuccessorsRequestInitial,
-        GetSuccessorsResponse,
+        GetSuccessorsResponse, Slicing,
     },
     with_state, with_state_mut,
 };
@@ -23,12 +23,20 @@ pub async fn heartbeat() {
     collect_metrics();
     maybe_burn_cycles();
 
-    if ingest_stable_blocks_into_utxoset() {
-        // Stable-block ingestion did work this round: it either finished a block
-        // or paused mid-block (time-sliced). Exit the heartbeat as a precaution to
-        // not exceed the instructions limit.
-        print("Ingested stable blocks (possibly paused mid-block). Exiting heartbeat.");
-        return;
+    match ingest_stable_blocks_into_utxoset() {
+        // Ingestion paused mid-block (time-sliced): a full budget was already spent, so
+        // exit the heartbeat as a precaution to not exceed the instructions limit.
+        Slicing::Paused(()) => {
+            print("Paused while ingesting a stable block. Exiting heartbeat.");
+            return;
+        }
+        // A stable block finished ingesting this round: exit for the same precaution.
+        Slicing::Done(true) => {
+            print("Done ingesting stable blocks. Exiting heartbeat.");
+            return;
+        }
+        // Nothing was ingested; carry on to fetching/processing.
+        Slicing::Done(false) => {}
     }
 
     if maybe_fetch_blocks().await {
@@ -193,7 +201,7 @@ async fn maybe_fetch_blocks() -> bool {
     true
 }
 
-fn ingest_stable_blocks_into_utxoset() -> bool {
+fn ingest_stable_blocks_into_utxoset() -> Slicing<(), bool> {
     with_state_mut(state::ingest_stable_blocks_into_utxoset)
 }
 
--- a/canister/src/heartbeat.rs
+++ b/canister/src/heartbeat.rs
@@ -774,6 +774,11 @@ mod test {
         );
         // ...and ingestion made real progress into the stable UTXO set.
         assert!(with_state(|s| s.utxos.next_height()) > 0);
+
+        // Restore the shared (thread-local) performance-counter mock to its default so this
+        // test does not affect others that may run later on the same thread.
+        runtime::set_performance_counter_step(0);
+        runtime::performance_counter_reset();
     }
 
     #[async_std::test]
--- a/canister/src/heartbeat.rs
+++ b/canister/src/heartbeat.rs
@@ -483,6 +483,12 @@ mod test {
         // Assert that the blocks have been ingested.
         assert_eq!(with_state(state::main_chain_height), 2);
 
+        // Number of `get_successors` calls issued so far (only the single fetch above).
+        // While stable blocks are still being ingested, the heartbeat must return early
+        // and must NOT issue a new `get_successors` request.
+        let fetches_after_processing =
+            runtime::GET_SUCCESSORS_RESPONSES_INDEX.with(|i| *i.borrow());
+
         // Ingest stable blocks.
         runtime::performance_counter_reset();
         heartbeat().await;
@@ -494,6 +500,12 @@ mod test {
         assert_eq!(partial_block.next_tx_idx, 2);
         assert_eq!(partial_block.next_input_idx, 1);
         assert_eq!(partial_block.next_output_idx, 0);
+        // The paused ingestion returned early without fetching new blocks.
+        assert_eq!(
+            runtime::GET_SUCCESSORS_RESPONSES_INDEX.with(|i| *i.borrow()),
+            fetches_after_processing,
+            "heartbeat fetched new blocks while a stable block was still being ingested",
+        );
 
         // Ingest more stable blocks.
         runtime::performance_counter_reset();
@@ -505,6 +517,12 @@ mod test {
         assert_eq!(partial_block.next_tx_idx, 5);
         assert_eq!(partial_block.next_input_idx, 1);
         assert_eq!(partial_block.next_output_idx, 0);
+        // Still ingesting, still no new fetch.
+        assert_eq!(
+            runtime::GET_SUCCESSORS_RESPONSES_INDEX.with(|i| *i.borrow()),
+            fetches_after_processing,
+            "heartbeat fetched new blocks while a stable block was still being ingested",
+        );
 
         // Only the genesis block has been fully processed, so the stable height is one.
         assert_eq!(with_state(|s| s.utxos.next_height()), 1);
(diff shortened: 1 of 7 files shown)


canister/src/state.rs
```diff
@@ -222,22 +222,35 @@ pub fn ingest_stable_blocks_into_utxoset(state: &mut State) -> bool {
...
-        &state.utxos.ingesting_block.clone(),
-    );
-    let has_state_changed = |state: &State| -> bool {
-        prev_state != (state.utxos.next_height(), &state.utxos.ingesting_block)
-    };
+    // Whether this call performed any stable-block ingestion work.
+    //
+    // Previously this was computed by cloning `ingesting_block` on entry and
+    // deep-comparing it (via `PartialEq`) afterwards. That clone + compare walks
+    // the whole `IngestingBlock`, including the accumulated `utxos_delta`, on
```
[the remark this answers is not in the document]
mbjorkqvist: Good catch — updated the rustdoc in aee1b5b to describe the actual contract (whether stable-block ingestion work was performed this call, including a paused time-sliced slice) instead of the stale “whether the state has changed” wording.

canister/src/state.rs
```diff
@@ -222,22 +222,35 @@ pub fn ingest_stable_blocks_into_utxoset(state: &mut State) -> bool {
...
+    // every heartbeat. With the deterministic memory tracker (DMT) charging
+    // instructions per accessed page, that per-slice cost grows with the delta
+    // and can consume the self-slicing budget, starving forward progress so that
+    // a block never finishes ingesting and the heartbeat never resumes fetching
+    // (DEFI-2954). We track a cheap bool instead.
+    //
+    // Returning `true` for every `Slicing::Paused` is equivalent to the old
+    // behaviour: each paused slice mutates `ingesting_block` (via
+    // `stats.num_rounds` / `ins_total`), so the old comparison was always `true`
```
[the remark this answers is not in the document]
mbjorkqvist: Done in aee1b5b — qualified it as `stats.ins_total` for consistency with `stats.num_rounds`.

canister/src/state.rs
```diff
@@ -610,6 +625,57 @@ mod test {
...
+        while state.utxos.ingesting_block.is_some() {
+            crate::runtime::performance_counter_reset();
+            assert!(ingest_stable_blocks_into_utxoset(&mut state));
+        }
+        assert_eq!(state.stable_height(), 2);
+
+        // Everything stable has been ingested -> no work -> false.
+        crate::runtime::performance_counter_reset();
+        assert!(!ingest_stable_blocks_into_utxoset(&mut state));
+    }
```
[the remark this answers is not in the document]
mbjorkqvist: Good point — added a reset of the performance-counter mock (step and counter back to 0) at the end of this test in a37a81d, so it can’t leak time-slicing state to other tests that run later on the same thread.

canister/src/heartbeat.rs
```diff
@@ -696,6 +696,86 @@ mod test {
...
+        );
+        // ...no heartbeat fetched while ingestion was ongoing (incl. the one that completed it)...
+        assert_eq!(
+            runtime::GET_SUCCESSORS_RESPONSES_INDEX.with(|i| *i.borrow()),
+            fetches_after_processing,
+            "heartbeat fetched new blocks while ingesting stable blocks",
+        );
+        // ...and ingestion made real progress into the stable UTXO set.
+        assert!(with_state(|s| s.utxos.next_height()) > 0);
+    }
```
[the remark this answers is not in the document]
mbjorkqvist: Fixed in a37a81d — the test now restores the step and counter to their defaults at the end, for the same isolation reason.

canister/src/state.rs
```diff
@@ -222,22 +223,35 @@ pub fn ingest_stable_blocks_into_utxoset(state: &mut State) -> bool {
...
     }
 
-    let prev_state = (
-        state.utxos.next_height(),
-        &state.utxos.ingesting_block.clone(),
-    );
-    let has_state_changed = |state: &State| -> bool {
-        prev_state != (state.utxos.next_height(), &state.utxos.ingesting_block)
-    };
+    // Whether this call performed any stable-block ingestion work.
```
gregorydemay: I would personally remove the whole comment, add it to the PR description if needed.

canister/src/state.rs
```diff
@@ -255,15 +269,16 @@ pub fn ingest_stable_blocks_into_utxoset(state: &mut State) -> bool {
...
                 state.metrics.block_ingestion_stats = stats;
-                pop_block(state, ingested_block_hash)
+                pop_block(state, ingested_block_hash);
+                did_work = true;
             }
         }
     }
 
-    has_state_changed(state)
+    did_work
```
gregorydemay: Here for a lack of a better place: Let's remove `Clone` from `IngestingBlock`, we can gate it for `cfg(test)` for the couple of tests that use it.

canister/src/state.rs
```diff
@@ -211,7 +211,8 @@ pub fn insert_block(state: &mut State, block: Block) -> Result<(), InsertBlockEr
 /// NOTE: This method does a form of time-slicing to stay within the instruction limit, and
 /// multiple calls may be required for all the stable blocks to be ingested.
 ///
-/// Returns a bool indicating whether or not the state has changed.
+/// Returns `true` if any stable-block ingestion work was performed during this call
+/// (including a paused, time-sliced slice), or `false` if there was nothing to ingest.
 pub fn ingest_stable_blocks_into_utxoset(state: &mut State) -> bool {
```
gregorydemay: I have the impression there is a mismatch in how the result is interpreted:
```
    if ingest_stable_blocks_into_utxoset() {
        // Exit the heartbeat if stable blocks had been ingested.
        // This is a precaution to not exceed the instructions limit.
        print("Done ingesting stable blocks.");
        return;
    }
```
From the log message it seems to be understood that ingesting stable blocks was completed, even if it was just paused. This was pre-existing behavior. Maybe adapt the print? Alternatively let `ingest_stable_blocks_into_utxoset` return a `Slicing<>` value?

canister/src/state.rs
```diff
@@ -610,6 +625,62 @@ mod test {
         assert_ne!(metrics_before, state.metrics.block_ingestion_stats);
     }
 
+    // The return value must be `true` whenever a stable block was ingested (fully or as a
+    // time-sliced chunk) and `false` when there was nothing to ingest. The heartbeat relies
+    // on this to decide whether to keep ingesting or move on to fetching new blocks.
+    #[test]
+    fn ingest_stable_blocks_into_utxoset_reports_whether_work_was_done() {
```
gregorydemay: I think we could fold it into the existing test `block_ingestion_stats_are_updated`

canister/src/heartbeat.rs
```diff
@@ -696,6 +696,91 @@ mod test {
         );
     }
 
+    // While a stable block is still being ingested, the heartbeat must return early (via
+    // `ingest_stable_blocks_into_utxoset`) and must NOT issue a new `get_successors` request.
+    #[async_std::test]
+    async fn heartbeat_does_not_fetch_while_ingesting_a_stable_block() {
```
gregorydemay: fold it into `time_slices_large_blocks`?

gregorydemay (commented):
Thanks @mbjorkqvist ! LGTM, some minor comments about Claude being a bit too eager to write code/comments :-)

canister/src/state.rs
mbjorkqvist: Done in `fbae76d` — trimmed it to a few lines keeping only the non-obvious behavioral note (why a paused slice counts as work), and moved the DMT/DEFI-2954 rationale to the PR description.

canister/src/state.rs
mbjorkqvist: Done in `81e7030` — `IngestingBlock` no longer derives `Clone` in production; it's now `#[cfg_attr(test, derive(Clone))]`, mirroring the existing `cfg(test)` `PartialEq` on `UtxoSet`.

canister/src/state.rs
mbjorkqvist: Good catch — went with the `Slicing<>` option in `a9e6a80`: `ingest_stable_blocks_into_utxoset` now returns `Slicing<(), bool>` — `Paused(())`, `Done(true)` (all stable blocks ingested this call) or `Done(false)` (nothing to ingest) — and the heartbeat matches on the variant and logs each case accurately. (`2280753` first just reworded the misleading print; `a9e6a80` supersedes it.)

canister/src/state.rs
mbjorkqvist: Done in `051d652` — folded the return-value assertions into that test and dropped the standalone one. Since it now also covers the return-value contract, I renamed it to `ingest_stable_blocks_updates_stats_and_reports_work` in `68b7c68`.

canister/src/heartbeat.rs
mbjorkqvist: Done in `292f265` — moved the "no `get_successors` fetch while ingesting" assertions into that test and removed the duplicate (also added an end-of-test mock reset for isolation). Renamed to `time_slices_large_blocks_without_fetching_while_ingesting` in `68b7c68`.

gregorydemay (approved):

PR #541, opened 2026-07-24
perf: DEFI-2957: Make BlockTree::get_hashes O(n) instead of O(n²)

mbjorkqvist:
## Purpose

Follow-up to DEFI-2954, targeting the Bitcoin canister's elevated steady-state cycle burn under the deterministic memory tracker (DMT).

The `heartbeat_steady_state_synced` benchmark (added in the base PR #540) showed that building the `get_successors` request dominated the synchronous per-round heartbeat work (~87%). The cause is `BlockTree::get_hashes`: it recursed by returning a fresh `Vec` per node and concatenating it into the parent, re-copying each node's descendants' hashes at every level — O(n²) on a long chain. The heartbeat rebuilds this list every round, so DMT charges it repeatedly.

This rewrites `get_hashes` as a single depth-first traversal into one shared accumulator (O(n)), preserving the pre-order / root-first ordering the heartbeat relies on (it pops index 0 as the anchor), and adds a test locking that contract.

## Measured effect

Via `heartbeat_steady_state_synced`:

| scope | before | after | Δ |
|---|---|---|---|
| `build_get_successors_request` | 707.6K | 18.0K | **−97.5%** |
| total synchronous round | 809.2K | 119.7K | **−85.2%** |

(`collect_metrics`/`tip_depths`, now the largest remaining term at ~74K, is a separate follow-up.)

## Review notes

- Stacked on #540 (the benchmark). Base will be retargeted to `master` once #540 merges; the `canbench_results.yml` diff here then shows the improvement cleanly.
- canbench measures instructions, not DMT page-access charges.

### DIFF
--- a/benchmarks/src/main.rs
+++ b/benchmarks/src/main.rs
@@ -1,9 +1,12 @@
 use bitcoin::consensus::Decodable;
 use bitcoin::constants::genesis_block;
 use bitcoin::{block::Header, consensus::Encodable, Block as BitcoinBlock};
-use canbench_rs::{bench, bench_fn, BenchResult};
+use canbench_rs::{bench, bench_fn, bench_scope, BenchResult};
 use ic_btc_canister::state::main_chain_height;
-use ic_btc_canister::{types::BlockHeaderBlob, with_state, with_state_mut};
+use ic_btc_canister::{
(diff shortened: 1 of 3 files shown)


ielashi (approved):
LGTM thanks @mbjorkqvist !

